PR #27, opened 2026-01-27
Move and update README files

antimora:
## Summary
- Moves `README.md` and `SUPPORTED-ONNX-OPS.md` from `crates/burn-onnx/` to repository root
- Adds README files for `burn-onnx` and `onnx-ir-derive` crates
- Updates all links to use absolute GitHub URLs (for proper rendering on crates.io)
- Updates each crate to use its own local README instead of workspace readme

## Test plan
- [x] `cargo check` passes
- [ ] Verify links work on GitHub


antimora:
@laggui merged without approval

PR #60, opened 2026-01-31
Add no-op node marking in onnx-ir processor for graph optimization

### LINKED ISSUE #12: Add no-op node marking in onnx-ir processor for graph optimization
### Summary

Enhance onnx-ir's NodeProcessor trait to allow nodes to signal they are no-ops during type inference. This enables automatic
elimination of identity operations without requiring burn-import to generate pass-through code.

### Motivation

Some ONNX operations become no-ops under certain conditions:

- Reshape: Reshape(scalar, [-1]) or Reshape(scalar, [1]) where input and output are both scalars
- Gather: Gather(scalar, 0) where gathering from a single-element scalar just returns that scalar
- Squeeze/Unsqueeze: When axes don't change the effective shape
- Identity: Always a no-op (currently handled via identity elimination pass)

Currently, these cases require burn-import to generate pass-through code like:
`let output = input;  // wasteful assignment`

With no-op marking, onnx-ir could automatically rewire the graph to skip these nodes entirely, similar to the existing identity elimination pass.

### Proposed Design

Add a method to NodeProcessor trait:

```rust
trait NodeProcessor {
    // ... existing methods ...

    /// Returns true if this node is a no-op (output equals input).
    /// When true, the node can be eliminated and its output rewired to its input.
    /// Called after type inference when input/output types are known.
    fn is_noop(&self, node: &RawNode) -> bool {
        false  // default: not a no-op
    }
}
```

### Benefits

1. Cleaner generated code: No unnecessary assignments
2. Unified optimization: Consolidates identity-like elimination in onnx-ir
3. Extensible: Easy to add no-op detection to other operators
4. Better debugging: Graph visualization shows actual computation flow

##3 Related

- https://github.com/tracel-ai/burn/pull/4146
- Current identity elimination pass in onnx-ir/src/ir/graph.rs

antimora:
## Summary

- Adds `is_noop()` method to `NodeProcessor` trait, allowing any processor to declare itself a no-op after type inference
- Generalizes the hardcoded Identity elimination in post-processing to use `is_noop()`, so any processor-declared no-op is automatically eliminated by rewiring consumers to the node's first input
- Implements `is_noop()` for Identity (always), Cast (same-type), Reshape (scalar-to-scalar), and Gather (scalar input)
- Adds integration tests with ONNX fixtures and unit tests for each `is_noop` implementation
- Updates `DEVELOPMENT-GUIDE.md` to document the new trait method

## Test plan

- [x] All existing `onnx-ir` unit tests pass (identity elimination unchanged behavior)
- [x] All existing `burn-onnx` tests pass
- [x] New unit tests for `is_noop()` on Cast, Reshape, Gather processors
- [x] New integration tests verifying Cast(F32->F32) is eliminated, Cast(F32->I64) is preserved, Identity still eliminated
- [x] Clean build with no warnings

Closes #12

### DIFF
--- a/crates/onnx-ir/src/phases/post_processing.rs
+++ b/crates/onnx-ir/src/phases/post_processing.rs
@@ -69,7 +69,7 @@ fn plan_noop_elimination(
         let node = &nodes[idx];
 
         if node.inputs.is_empty() {
-            log::warn!("Identity node {} has no inputs, skipping", node.name);
+            log::warn!("No-op node {} has no inputs, skipping", node.name);
             continue;
         }
 
@@ -83,10 +83,10 @@ fn plan_noop_elimination(
         // the original ONNX names. We need to find the original name by reverse-looking up
         // the node_output_map (which maps original names to node indices).
         //
-        // For this Identity node at index `idx` with output index 0, find the original name.
+        // For this no-op node at index `idx` with output index 0, find the original name.
         for (original_name, &(node_idx, output_idx)) in node_output_map.iter() {
             if node_idx == idx && output_idx == 0 {
-                // Found the original ONNX output name for this Identity node
+                // Found the original ONNX output name for this no-op node
                 if original_name != output_name {
                     rewire_map.insert(original_name.clone(), input_name.clone());
                 }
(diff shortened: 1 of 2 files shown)


antimora:
All three Copilot review comments addressed in 8a3475b:

1. Updated stale "Identity node" references in comments/log messages to "no-op node"
2. Updated constant lifting comment and log message to say "no-op elimination"
3. Removed unused `numpy` import and dependency from generate_noop.py

PR #85, opened 2026-02-02
Preserve concrete dimensions in static_shape when some dims are symbolic

### LINKED ISSUE #62: Bug related to symbolic dimensions
I got this bug while working with onnx in a agentic coding cli:

TensorShapeProto conversion silently drops symbolic dimensions
Problem:
When converting ONNX TensorShapeProto to Vec, symbolic dimensions (like batch size N) are silently skipped. For a model with input shape [N, 3, 224, 224], the resulting static_shape becomes None (or an incomplete vector), even though 3 of the 4 dimensions are concrete and usable.
Location:
crates/onnx-ir/src/proto_conversion.rs, lines 508-520:
impl TryFrom for Vec {
type Error = ParseError;
fn try_from(shape: TensorShapeProto) -> Result<Vec, Self::Error> {
let mut result = Vec::new();
for dim in shape.dim {
if let Value::DimValue(value) = dim.value.unwrap() {
result.push(value as usize);
}
// ⚠️ DimParam (symbolic) dimensions are silently ignored
}
Ok(result)
}
}
Impact:

Most ONNX models have symbolic batch dimensions (standard practice)
Downstream users cannot determine any shape info, even for concrete dimensions
Users must preprocess models to make all dimensions concrete
Suggested fix:
Option A - Return error if any dimension is symbolic:
for dim in shape.dim {
match dim.value {
Some(Value::DimValue(value)) => result.push(value as usize),
Some(Value::DimParam(name)) => {
return Err(ParseError::SymbolicDimension(name));
}
None => return Err(ParseError::UnknownDimension),
}
}
Option B - Represent symbolic dims as sentinel (e.g., -1 or enum)
Option C - Add partial_shape: Vec alongside static_shape: Option
Workaround (current fix for users):
import onnx
model = onnx.load("model.onnx")
for inp in model.graph.input:
if inp.type.tensor_type.shape.dim[0].HasField('dim_param'):
inp.type.tensor_type.shape.dim[0].dim_value = 1
inp.type.tensor_type.shape.dim[0].ClearField('dim_param')
onnx.save(model, "model_concrete.onnx")

antimora:
## Summary
- Changes `static_shape` from `Option<Vec<usize>>` to `Option<Vec<Option<usize>>>` so each dimension is independently `Some(val)` or `None` (symbolic)
- Previously, if ANY dimension was symbolic (e.g. batch dim `N`), all concrete shape info was discarded. Now `[N, 3, 224, 224]` produces `Some([None, Some(3), Some(224), Some(224)])`
- Adds `static_shape_known()` convenience method returning `Option<Vec<usize>>` only when all dims are concrete, allowing ~80% of call sites to migrate with zero logic changes

Fixes #62

## Test plan
- [x] `cargo test -p onnx-ir` (35 passed)
- [x] `cargo test -p burn-onnx` (10 passed)
- [x] `cargo test -p onnx-tests` (417 passed)
- [x] Reproduction test asserts `[N, 3, 224, 224]` preserves concrete dims as `Some([None, Some(3), Some(224), Some(224)])`


antimora:
cargo xtask validate run locally and it's passing

PR #92, opened 2026-02-06
fix panic when converting fp16 model

gorynich2048:
Got `Unsupported tensor type` panic when tried to convert ISNet and RMBG-1.4 fp16 ONNX models.
After this fix they converted and worked.


antimora:
Superseded by #95 which extends this fix to handle all DType variants. Thank you @gorynich2048 for identifying and initially fixing this issue!

I couldn't push the changes direct to your branch because you checked out on main in your repo. I had fix a couple issues.

PR #103, opened 2026-02-06
fix: merge static_shape during type inference instead of replacing

### LINKED ISSUE #100: Make static_shape propagation more robust by merging with value_info
## Problem

During type inference, when syncing inferred output types to downstream inputs, the current implementation **replaces** the input type entirely. This means if the ONNX proto's `value_info` provided accurate partial shape info (e.g., `[dynamic, 1]`), but an upstream processor set `static_shape: None`, the valuable info from `value_info` gets overwritten with less-informative data.

This makes `static_shape` propagation fragile: one careless `static_shape: None` anywhere in the chain breaks downstream consumers that need partial shape info (e.g., GatherND needs the last dimension of indices to compute output rank).

## Proposed Solution

When the type inference loop syncs an inferred type to a downstream input, **merge** the `static_shape` rather than replacing it. Take the more informative value for each dimension:

- If value_info says `Some(vec![None, Some(1)])` and inferred type says `static_shape: None`, keep `Some(vec![None, Some(1)])`
- If value_info says `Some(vec![None, Some(1)])` and inferred type says `Some(vec![Some(5), None])`, merge to `Some(vec![Some(5), Some(1)])`
- If both agree, no change needed

This applies to the sync logic in `crates/onnx-ir/src/phases/type_inference.rs` (both intra-iteration sync and cross-iteration sync).

## Context

Discovered while implementing GatherND (#97). The real-world model `meiki.text.detect.tiny.v0.onnx` has a `NonZero -> Transpose -> GatherND` chain where NonZero was setting `static_shape: None`, which propagated through Transpose and caused GatherND to fail. While we fixed both NonZero and Transpose individually (#99), the underlying architecture remains fragile.

antimora:
## Summary

- Fixes fragile `static_shape` propagation during type inference by merging shape info instead of replacing it entirely
- When syncing inferred output types to downstream inputs, preserves valuable partial shape info from `value_info` that would otherwise be lost when an upstream processor sets `static_shape: None`
- Adds `TensorType::merge_static_shape()` and `ArgType::merge_static_shape()` methods with clear merge rules

## Merge Rules

- If inferred type has no shape info but `value_info` does, use `value_info`'s shape
- If both have shapes with matching rank, merge dimension by dimension (prefer known values over unknown)
- Inferred type's `dtype`/`rank` always take precedence

## Test plan

- [x] Added 8 unit tests for merge behavior in `argument.rs`
- [x] Added 2 integration tests for type inference merge in `type_inference.rs`
- [x] All existing tests pass (`cargo xtask validate`)

Closes #100

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


antimora:
All three issues have been addressed in the latest commit:

1. **Convergence issue (lines 150, 175)**: Removed the `else if shape_changed` branches. The `input.ty != merged_type` check is sufficient to detect actual state changes.

2. **Rank check (line 153)**: Added an explicit `self.rank != other.rank` guard at the top of `merge_static_shape()` to ensure ranks match before any merge attempt, consistent with the documentation.

PR #198, opened 2026-02-21
Fix #193: avoid Gather scalar round-trip and make xtask validate find typos

### LINKED ISSUE #193: GPU-CPU round-trips in ops that need native scalars/shapes from tensors
## Summary

Several operators force GPU-to-CPU round-trips because Burn's APIs accept native Rust types (scalars, `[usize]` shapes) rather than tensors. These are **not regressions** from #191 (ScalarTensor) -- they are pre-existing patterns, same count before and after the change.

## Burn framework limitations (not actionable in burn-onnx)

These require upstream Burn API changes to fix:

### GatherND / ScatterND full index round-trip (GPU -> CPU -> GPU)
- **Files:** `gather_nd.rs`, `scatter_nd.rs`
- **Impact:** High
- **Root cause:** No native `gather_nd` / `scatter_nd` kernel in Burn. The implementation pulls index tensors to CPU to decompose into per-element `select()` calls.

### Resize pulls sizes/scales to CPU
- **Files:** `resize.rs`
- **Impact:** Medium
- **Root cause:** Burn's `interpolate` takes scalar parameters (output size as integers), not tensors.

### Reshape / Expand / Unsqueeze runtime shape tensors to CPU
- **Files:** `reshape.rs`, `expand.rs`, `unsqueeze.rs`
- **Impact:** Low
- **Root cause:** Burn's `reshape` takes `[usize]`, not a shape tensor. Same for `expand` and `unsqueeze` with dynamic shapes.

### Gather scalar-index round-trip
- **File:** `gather.rs:193`
- **Impact:** Low
- **Pattern:** `from_data([idx]) -> select() -> into_scalar()` -- creates a 1-element tensor on GPU, selects, then pulls result back to CPU.
- **Possible fix:** Keep the result as `ScalarTensor` instead of `ScalarNative`, avoiding the round-trip entirely. Requires checking whether downstream consumers can accept `ScalarTensor`.

amyssnippet:
### Checklist

This PR fixes the actionable part of #193 by removing the temporary index tensor round-trip in Gather scalar-native codegen and adds an xtask PATH fallback so cargo xtask validate can locate typos after installation even when ~/.cargo/bin is not on PATH.

fixes #193

### DIFF
--- a/crates/burn-onnx/src/burn/node/gather.rs
+++ b/crates/burn-onnx/src/burn/node/gather.rs
@@ -227,8 +227,10 @@ fn forward_tensor_gather(
                         })
                         .collect::<Vec<_>>();
                     quote! {
-                        #pre_convert
-                        let #output = #input.slice(s![#(#slice_args),*]);
+                        let #output = {
+                            #pre_convert
+                            #input.slice(s![#(#slice_args),*])
+                        };
                     }
                 }
                 _ => panic!(
@@ -598,8 +600,10 @@ mod tests {
             values: Tensor<B, 1>,
             position: Tensor<B, 1, Int>,
         ) -> Tensor<B, 1> {
-            let __scalar_idx = position.into_scalar().elem::<i64>();
-            let result = values.slice(s![__scalar_idx]);
+            let result = {
+                let __scalar_idx = position.into_scalar().elem::<i64>();
+                values.slice(s![__scalar_idx])
+            };
             result
         }
         ");


xtask/src/main.rs
```diff
@@ -32,7 +32,36 @@ pub enum Command {
     ModelCheck(model_check::ModelCheckArgs),
 }
 
+fn ensure_cargo_bin_on_path() {
+    let cargo_home = env::var_os("CARGO_HOME")
+        .map(PathBuf::from)
+        .or_else(|| env::var_os("HOME").map(|home| PathBuf::from(home).join(".cargo")));
```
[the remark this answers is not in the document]
amyssnippet: added this one

amyssnippet:
@antimora it seems everything is perfect, kindly have a review, rest i can see it is mergable!

crates/burn-onnx/src/burn/node/gather.rs
```diff
@@ -240,17 +240,21 @@ fn forward_tensor_gather(
...
             match &index_arg.ty {
                 ArgType::ScalarNative(_) | ArgType::ScalarTensor(_) => {
-                    let index = if index_arg.ty.is_scalar_tensor() {
-                        let tensor = scope.arg(index_arg);
-                        on_device_to_native(quote! { #tensor }, &index_arg.ty.elem_type())
-                    } else {
-                        let ident = arg_to_ident(index_arg);
-                        quote! { #ident }
-                    };
+                    let (pre_convert, index_ident) = scalar_index_to_ident(index_arg, scope);
```
antimora, line 247: Nit: the `ScalarTensor` path above (lines 214-230) still uses the old `Tensor::select` + device-allocated index tensor approach. Since type inference guarantees `input_rank == 1` for ScalarTensor output, `input.slice(s![idx])` would work there too and avoid the device allocation. Not a bug (both are correct), but could unify the two scalar paths for consistency.

amyssnippet: done

crates/burn-onnx/src/burn/node/gather.rs
```diff
@@ -215,18 +215,20 @@ fn forward_tensor_gather(
...
+                            }
+                        })
+                        .collect::<Vec<_>>();
                     quote! {
-                        let #output = {
-                            let indices = Tensor::<B, 1, _>::from_data([#index], &*self.device);
-                            Tensor::select(#input, #dim, indices)
-                        };
+                        #pre_convert
+                        let #output = #input.slice(s![#(#slice_args),*]);
```
antimora: `#pre_convert` should be scoped inside `let #output = {}` not to pollute global name space.

amyssnippet: done

Issue #158, opened 2026-02-12
Implement Shrink operator


antimora:
## Context

Gap analysis against [tract](https://github.com/sonos/tract).

## Operator

**Shrink** - applies hard or soft shrinkage to input tensor.

- [ONNX spec](https://onnx.ai/onnx/operators/onnx__Shrink.html)
- Currently listed as `unsupported::ShrinkNode` in `node.rs`

## Attributes
- `bias` (float, default 0.0)
- `lambd` (float, default 0.5)

## Formula
```
If x < -lambd: y = x + bias
If x > lambd:  y = x - bias
Otherwise:     y = 0
```

aditya0by0:
@antimora, for this issue, does one needs to implement two separate activation functions in Burn:activation: **Softshrink** and **Hardshrink** first, before the implementation for this issue?

In PyTorch, these are provided as two distinct activation modules:

* PyTorch Softshrink: [https://docs.pytorch.org/docs/stable/generated/torch.nn.Softshrink.html#torch.nn.Softshrink](https://docs.pytorch.org/docs/stable/generated/torch.nn.Softshrink.html#torch.nn.Softshrink)
* PyTorch Hardshrink: [https://docs.pytorch.org/docs/stable/generated/torch.nn.Hardshrink.html#torch.nn.Hardshrink](https://docs.pytorch.org/docs/stable/generated/torch.nn.Hardshrink.html#torch.nn.Hardshrink)

Since ONNX defines both behaviors under a single **Shrink** operator (controlled by the `bias` attribute), I wanted to confirm whether we should expose them as two separate activation functions in Burn (similar to PyTorch), or implement a single Shrink activation with configurable parameters.

antimora:
> [@antimora](https://github.com/antimora), for this issue, does one needs to implement two separate activation functions in Burn:activation: **Softshrink** and **Hardshrink** first, before the implementation for this issue?
> 
> In PyTorch, these are provided as two distinct activation modules:
> 
> * PyTorch Softshrink: https://docs.pytorch.org/docs/stable/generated/torch.nn.Softshrink.html#torch.nn.Softshrink
> * PyTorch Hardshrink: https://docs.pytorch.org/docs/stable/generated/torch.nn.Hardshrink.html#torch.nn.Hardshrink
> 
[...]

PR #200, opened 2026-02-21
Implement Shrink ONNX operator

aditya0by0:
### Checklist

### Related Issues/PRs

Closes: https://github.com/tracel-ai/burn-onnx/issues/158

### Changes

_Summarize the problem being addressed and your solution._

### Testing

_Describe how these changes have been tested._

### DIFF
--- a/crates/onnx-ir/src/node/shrink.rs
+++ b/crates/onnx-ir/src/node/shrink.rs
@@ -8,7 +8,7 @@
 //! - **Opset 9**: Initial version
 use onnx_ir_derive::NodeBuilder;
 
-use crate::ir::{ArgType, Argument, Node, RawNode};
+use crate::ir::{Argument, Node, RawNode};
 use crate::processor::{
     InputSpec, NodeProcessor, NodeSpec, OutputPreferences, OutputSpec, ProcessError,
 };
@@ -51,21 +51,7 @@ impl NodeProcessor for ShrinkProcessor {
         _opset: usize,
         _output_preferences: &OutputPreferences,
     ) -> Result<(), ProcessError> {
-        // Get input tensor type
-        let input_tensor = match &node.inputs[0].ty {
-            ArgType::Tensor(tensor) => tensor,
-            _ => {
-                return Err(ProcessError::TypeMismatch {
-                    expected: "Tensor".to_string(),
-                    actual: format!("{:?}", node.inputs[0].ty),
-                });
-            }
-        };
-
-        // Output has the same type and shape as input
-        node.outputs[0].ty = ArgType::Tensor(input_tensor.clone());
-
-        let _config = self.extract_config(node, _opset)?;
+        crate::processor::same_as_input(node);
         Ok(())
     }
 
@@ -81,12 +67,7 @@ impl NodeProcessor for ShrinkProcessor {
                 "lambd" => {
                     lambda = value.clone().into_f32() as f64;
                 }
-                _ => {
-                    return Err(ProcessError::InvalidAttribute {
-                        name: key.clone(),
-                        reason: format!("Unexpected attribute for Shrink: {}", key),
-                    });
-                }
+                _ => {}
             }
         }
 


antimora:
Merged the latest from burn https://github.com/tracel-ai/burn-onnx/pull/223 so your shrink functions should be available now

crates/onnx-tests/tests/shrink/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+            [[-2.0, 0.0, 0.0, 0.0, 2.0], [-3.0, -2.5, 0.0, 0.0, 3.0]],
+            &device,
+        );
+        let expected_with_bias = Tensor::<TestBackend, 2>::from_floats(
+            [[-0.5, 0.0, 0.0, 0.0, 0.5], [-1.5, -1.0, 0.0, 0.0, 1.5]],
+            &device,
+        );
+
+        assert_eq!(out_no_bias.into_data(), expected_no_bias.into_data());
+        assert_eq!(out_with_bias.into_data(), expected_with_bias.into_data());
```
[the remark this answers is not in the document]
aditya0by0: will this really cause a problem ?

crates/onnx-ir/src/node/shrink.rs
```diff
@@ -0,0 +1,170 @@
...
+            _ => {
+                return Err(ProcessError::TypeMismatch {
+                    expected: "Tensor".to_string(),
+                    actual: format!("{:?}", node.inputs[0].ty),
+                });
+            }
+        };
+
+        // Output has the same type and shape as input
+        node.outputs[0].ty = ArgType::Tensor(input_tensor.clone());
```
antimora: This whole manual match + `TypeMismatch` block can be replaced with:
```rust
crate::processor::same_as_input(node);
```
Shrink output has the same type and shape as input, which is exactly what `same_as_input` does. The `TypeMismatch` guard is also not needed since the framework validates input types upstream. See `EluProcessor::infer_types` for reference.

crates/onnx-ir/src/node/shrink.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let _config = self.extract_config(node, _opset)?;
```
antimora: This `extract_config` call is unused (result assigned to `_config`). If the intent was early attribute validation, that should happen inline here. But once we drop the unknown-attribute rejection in `extract_config` (see comment below), this call serves no purpose and can be removed.

crates/onnx-ir/src/node/shrink.rs
```diff
@@ -0,0 +1,170 @@
...
+                }
+                "lambd" => {
+                    lambda = value.clone().into_f32() as f64;
+                }
+                _ => {
+                    return Err(ProcessError::InvalidAttribute {
+                        name: key.clone(),
+                        reason: format!("Unexpected attribute for Shrink: {}", key),
+                    });
+                }
```
antimora: The `_ => return Err(...)` arm rejects unknown attributes, but DEVELOPMENT-GUIDE.md says:

> Processors should extract the attributes they need and ignore the rest. Do not iterate over all attributes to reject unknown ones, as ONNX may add new attributes in future opsets.

Just extract `"bias"` and `"lambd"` and ignore the rest:
```rust
for (key, value) in node.attrs.iter() {
    match key.as_str() {
        "bias" => bias = value.clone().into_f32() as f64,
        "lambd" => lambda = value.clone().into_f32() as f64,
        _ => {}
    }
}
```
(I know a bunch of existing ops have this same pattern, that's a pre-existing issue. No need to carry it forward though.)

crates/onnx-tests/tests/shrink/mod.rs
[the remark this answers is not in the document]
antimora: Exact equality is fine here. Shrink is pure conditional logic (compare against lambda, then add/subtract bias or zero out). There's no exp/log/sqrt or any accumulated floating point error path. The results are either exactly `x + bias`, `x - bias`, or `0.0`. Other piecewise-linear ops like Relu and LeakyRelu use exact equality in their tests too.

antimora (changes requested):

PR #233, opened 2026-03-01
Add RTMW3D model check

antimora:
## Summary
- Add model check for RTMW3D-x (3D human pose estimation from mmpose)
- Downloads from HuggingFace (`Soykaf/RTMW3D-x`), processes to opset 16, validates inference against onnxruntime reference
- Model: 550 nodes, input `[1, 3, 384, 288]`, 3 outputs (133 keypoints)
- Build succeeds (codegen + compilation pass)

Ref: tracel-ai/models#74

## Test plan
- [x] `get_model.py` downloads model and generates test data
- [x] `cargo build -p burn-onnx-model-checks-rtmw3d` compiles successfully
- [ ] `cargo run -p burn-onnx-model-checks-rtmw3d` passes inference validation

### DIFF
--- a/crates/model-checks/rtmw3d/build.rs
+++ b/crates/model-checks/rtmw3d/build.rs
@@ -21,7 +21,7 @@ fn main() {
         eprintln!("Please run the following command to download and prepare the model:");
         eprintln!("  uv run get_model.py");
         eprintln!();
-        eprintln!("This will download the RTMW3D-x model and convert it to ONNX format.");
+        eprintln!("This will download the RTMW3D-x ONNX model and prepare it for codegen.");
         std::process::exit(1);
     }
 


antimora:
## Copilot review responses

**Comments 1 & 2 (TestData keys mismatch):** Not applicable. The actual ONNX output names for this model are `output`, `1554`, `1556` (not `0`, `1`, `2`). The first output is literally named `output`, so it is saved under key `output` in the .pt file, matching the Rust `TestData.output` field exactly. The numeric outputs are saved as `output_1554`/`output_1556` but are not loaded in TestData since we only validate the primary output.

**Comment 3 (misleading build.rs message):** Fixed in 8593cf10.

PR #267, opened 2026-03-24
Raise error on missing `BitShift` direction attribute

### LINKED ISSUE #215: BitShift: required `direction` attribute silently defaults to "left"
## Summary

The ONNX spec marks the `direction` attribute on BitShift as **required**, but the implementation silently defaults to `"left"` when it's missing. This means a malformed ONNX model with no `direction` attribute would be accepted instead of rejected.

## Location

`crates/onnx-ir/src/node/bitshift.rs`, lines 83 and 94

## Suggested fix

In `extract_config`, treat a missing `direction` as an error (`ProcessError::MissingAttribute`) rather than defaulting to `"left"`. The `Direction` enum already exists, so the fix is just changing the `unwrap_or_else` to an error.

christeefy:
### Checklist

- [ ] ~Confirm relevant documentation has been updated.~
- [ ] ~For new/modified ONNX operators: verified implementation matches `onnx-spec/ops/<OpName>.md`.~

### Related Issues/PRs

Closes #215

### Changes

As per the linked issue,
- raised `ProcessError::MissingAttribute` instead of defaulting it to `"left"` when the `direction` attribute is unspecified.
- update unit test

### Testing

Updated the final unit test to raise an error.

### DIFF
--- a/crates/onnx-ir/src/node/bitshift.rs
+++ b/crates/onnx-ir/src/node/bitshift.rs
@@ -76,11 +76,11 @@ impl NodeProcessor for BitShiftProcessor {
     fn infer_types(
         &self,
         node: &mut RawNode,
-        _opset: usize,
+        opset: usize,
         _output_preferences: &OutputPreferences,
     ) -> Result<(), ProcessError> {
         // Validate node attributes
-        let _ = self.extract_config(node, _opset)?;
+        let _ = self.extract_config(node, opset)?;
 
         // Output type is same as input with broadcasting
         crate::processor::same_as_input_broadcast(node);


christeefy:
Note: 
I ran `cargo xtask validate` but it failed on two separate occasions. 

1. `cargo audit` surfaced 4 vulnerabilities, which was fixed via `cargo update`. I'll check in the changes tentatively, albeit it being rather large. Let me know if you would prefer a more targeted `cargo update`.
2. However, it also failed on the Workspace Unit Tests step, where I wasn't able to compile some inline asm of `cortex-m` and `pro-proc` (I'm on a Mac). Please advise what I should do, as I'm unfamiliar with the setup. That said, `cargo nextest` ran successfully.

crates/onnx-ir/src/node/bitshift.rs
```diff
@@ -91,12 +91,11 @@ impl NodeProcessor for BitShiftProcessor {
 
     fn extract_config(&self, node: &RawNode, _opset: usize) -> Result<Self::Config, ProcessError> {
         // Extract direction attribute
-        // FIXME: Spec marks 'direction' as required, but we provide default "left"
         let direction_str = node
             .attrs
             .get("direction")
             .map(|val| val.clone().into_string())
-            .unwrap_or_else(|| "left".to_string());
+            .ok_or_else(|| ProcessError::MissingAttribute("direction".to_string()))?;
```
antimora, line 97: Now that this properly returns an error for missing `direction`, there's a stale FIXME a few lines above (lines 82-84) that still says `extract_config` provides a default `"left"`:

```rust
// TODO: Add validation for unexpected attributes
// FIXME: Spec says 'direction' is required but extract_config provides default "left"
// Should either validate presence here or update spec documentation
```

The FIXME is just wrong now. And the TODO conflicts with project policy ("don't reject unknown ONNX attributes"). I'd remove all three lines.

Also worth noting: `infer_types()` doesn't validate `direction` at all, so the error only surfaces as a panic via `.expect()` in `build_node()` (line 113). Adding a quick check in `infer_types` would catch this in the phase that properly propagates `Result`:

```rust
node.attrs.get("direction")
    .ok_or_else(|| ProcessError::MissingAttribute("direction".to_string()))?;
```

That way the user gets a clean `ProcessError` instead of a panic, which is the whole point of this PR.

christeefy: Thanks for the pointing that out; missed it as I'm still familiarizing with this codebase.

I've implemented your suggestion with a call to `.extract_config` instead, consistent to other op's impls, like in `concat.rs` and [`slice.rs`](https://github.com/christeefy/burn-onnx/blob/2990ceddd8a81ec5e8ecdcce98b541f24662082b/crates/onnx-ir/src/node/slice.rs#L170).

That way, missing and invalid directions are both raised earlier in the pipeline's type inference phase, with a minimal and hopefully negligible performance cost. 

PTAL!

christeefy:
Thanks @antimora! Love to contribute more when I have spare cycles, though I'll need additional guidance as I get up to speed.

PR #286, opened 2026-04-02
Implement LRN operator

### LINKED ISSUE #154: Implement LRN (Local Response Normalization) operator
## Context

LRN is the single operator blocking 2 of 9 classic ONNX model zoo models (AlexNet, ZFNet).

## Operator

**LRN** (Local Response Normalization) - normalizes over local input regions. Used in older CNN architectures.

- [ONNX spec](https://onnx.ai/onnx/operators/onnx__LRN.html)
- Currently listed as `unsupported::LrnNode` in `node.rs`

## Attributes
- `alpha` (float, default 0.0001)
- `beta` (float, default 0.75)
- `bias` (float, default 1.0)
- `size` (int, required) - number of channels to sum over

## Models unblocked
- bvlc_alexnet
- zfnet512

## Priority
High - simple formula, unblocks classic model zoo models.

christeefy:
### Checklist

### Related Issues/PRs

Closes #154.

### Changes

Added support for the LRN operator, by adding relevant scaffolding for the ONNX-IR and Burn codegen portions, as well as integration tests in the `onnx-tests` crate.

### Testing

- Unit tests in `onnx-ir`
- Snapshot tests in `burn-onnx`
- Integration tests in `onnx-tests`


crates/burn-onnx/src/burn/node/lrn.rs
```diff
@@ -0,0 +1,87 @@
...
+            for c in 0..num_channels {
+                let win_start = if c >= pad_left { c - pad_left } else { 0 };
+                let win_end = (c + pad_right + 1).min(num_channels); // exclusive
+                let win_len = win_end - win_start;
+                // Sum over the channel window [win_start, win_end) using narrow
+                let window_sum = squared.clone().narrow(1, win_start, win_len).sum_dim(1);
+                square_sum = square_sum.slice_assign([0..shape[0], c..c + 1], window_sum);
+            }
+
+            let #output = #input / (#bias + #alpha / #size as f32 * square_sum).powf_scalar(#beta);
```
christeefy, line 49: I went with a naive implementation here as a first-pass. It should be possible to make this significantly faster (on accelerated hardware) and less memory-intensive by reimplementing it as a 1D convolution.

Is that something you would be interested in a future PR? Is there priority to include it as a native operator in the `burn` repo?

antimora: At very least least we should file an issue with Burn and add a comment with the link. 

Direct burn users can also benefit from this op, as it appears there is one in Pytorch: https://docs.pytorch.org/docs/stable/generated/torch.nn.LocalResponseNorm.html 

This Op is mostly for wide support of models (older models seem to use this).

crates/onnx-ir/src/node/lrn.rs
```diff
@@ -0,0 +1,352 @@
...
+            .ok_or_else(|| ProcessError::MissingInput("Missing input".to_string()))?;
+        let ArgType::Tensor(ref tensor_ty) = arg.ty else {
+            return Err(ProcessError::TypeMismatch {
+                expected: "Input should be a tensor".to_string(),
+                actual: format!("{:?}", arg.ty),
+            });
+        };
+        if tensor_ty.rank < 4 {
+            return Err(ProcessError::TypeMismatch {
+                expected: "Expecting a tensor of at least rank 4".to_string(),
```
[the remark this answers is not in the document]
antimora: Valid. The ONNX spec defines the input shape as `(N x C x D1 x D2, ..., Dk)`, so the minimum rank is 3 (N, C, D1), not 4. The codegen snapshot test also uses rank-3 tensors, which is inconsistent with the rank-4 check. The unit test `test_lrn_rejects_low_rank_tensor` also incorrectly includes rank 3 as a rejected case.

crates/onnx-ir/src/node/lrn.rs
```diff
@@ -0,0 +1,352 @@
...
+            .map(|val| val.clone().into_f32())
+            .unwrap_or(BIAS_DEFAULT);
+
+        // Validate that `size` exists, and is of type `int`
+        let size = node
+            .attrs
+            .get("size")
+            .map(|val| val.clone().into_i64())
+            .ok_or_else(|| ProcessError::MissingAttribute("size".to_string()))?;
+
```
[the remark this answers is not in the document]
antimora: Partially valid. The suggestion to validate `size > 0` is reasonable since a zero or negative window size is mathematically nonsensical (would cause division by zero in `alpha / size`). However, the existing code already correctly validates that `size` exists and returns `MissingAttribute` when absent, so the first part of the suggestion is redundant. Only the `size > 0` guard is new and useful.

crates/onnx-ir/src/node/lrn.rs
```diff
@@ -0,0 +1,352 @@
+//! # LRN (Local Response Normalization)
+//!
+//! Local Response Normalization as proposed in the AlexNet paper. Normalizes
+//! over local input regions across the channel dimension.
+//!
+//! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__Lrn.html>
```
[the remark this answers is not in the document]
antimora: Valid. Confirmed that `onnx__Lrn.html` returns a 404 while `onnx__LRN.html` returns 200. The URL is case-sensitive and needs to be corrected.

crates/onnx-ir/src/node/lrn.rs
```diff
@@ -0,0 +1,352 @@
...
+            return Err(ProcessError::TypeMismatch {
+                expected: "Only BF16, F16, F32, F64 tensor dtypes are supported".to_string(),
+                actual: format!("{:?}", tensor_ty.dtype),
+            });
+        } else if !matches!(tensor_ty.dtype, DType::F16 | DType::F32 | DType::F64) {
+            return Err(ProcessError::TypeMismatch {
+                expected: "Only F16, F32, F64 tensor dtypes are supported".to_string(),
+                actual: format!("{:?}", tensor_ty.dtype),
+            });
+        }
```
antimora: This `if/else if` has a logic bug that rejects BF16 at opset >= 13, which is exactly when BF16 was added.

Walk through it with `opset=13, dtype=BF16`:
1. First branch: `opset >= 13 && !matches!(BF16, BF16|F16|F32|F64)` = `true && false` = `false`
2. Falls to `else if`: `!matches!(BF16, F16|F32|F64)` = `true` => BF16 rejected

Needs to be restructured so the two opset paths are independent:

```rust
if opset >= 13 {
    if !matches!(tensor_ty.dtype, DType::BF16 | DType::F16 | DType::F32 | DType::F64) {
        return Err(ProcessError::TypeMismatch { ... });
    }
} else if !matches!(tensor_ty.dtype, DType::F16 | DType::F32 | DType::F64) {
    return Err(ProcessError::TypeMismatch { ... });
}
```

There's no positive test for BF16-at-opset-13 either, which is why this slipped through. Adding something like `test_lrn_accepts_bfloat_at_opset_13` would catch it.

crates/onnx-tests/tests/lrn/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+                [0.19686124, 0.73846656, 0.17136829],
+                [-0.11564828, -0.30110368, -1.47852200],
+            ],
+        ]]);
+
+        output
+            .to_data()
+            .assert_approx_eq::<FT>(&expected, Tolerance::default());
+    }
+}
```
antimora, line 173: Both test models use the ONNX defaults for alpha/beta/bias (0.0001, 0.75, 1.0). Because alpha is so small, most output values are nearly identical to the input, which means a broken formula could still pass within tolerance for most elements.

Worth adding a model with non-default parameters (e.g. larger alpha, different beta) so the normalization effect is actually visible in the output. The project convention is to cover at least one non-default configuration per operator.

crates/burn-onnx/src/burn/node/lrn.rs
```diff
@@ -0,0 +1,87 @@
...
+                let win_len = win_end - win_start;
+                let window_sum = squared.clone().narrow(1, win_start, win_len).sum_dim(1);
+                square_sum = square_sum.slice_assign([0..shape[0], c..c + 1], window_sum);
+            }
+            let output = input / (1f32 + 0.0001f32 / 2 as f32 * square_sum).powf_scalar(0.75f32);
+            output
+        }
+        ");
+    }
+}
```
antimora, line 120: Only one snapshot test here. A second one with different alpha/beta/bias/size values would verify the float literal rendering and show the generated code actually changes with different configs.

crates/onnx-tests/tests/lrn/lrn.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env -S uv run --script
+
+# /// script
+# dependencies = [
+#   "numpy==2.2.4",
+#   "onnx==1.19.0",
+# ]
+# ///
+
+# used to generate model: lrn.onnx
```
antimora: Nit: comment says "used to generate model: lrn.onnx" but the script generates `lr_size3.onnx` and `lr_size2.onnx`. Also, there's a `lrn.onnx` file in this directory that isn't registered in `build.rs`, looks like a leftover.

crates/onnx-ir/src/node/lrn.rs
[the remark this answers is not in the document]
christeefy: Great catch!

crates/onnx-ir/src/node/lrn.rs
christeefy: Ah yes, you're right. Thanks for pointing that out.

christeefy:
Addressed most of the comments here. The expected values generated by the `ReferenceEvaluator` look a bit sus — only the first slice of the tensors change with `alpha` is varied. Will look more deeply into it before requesting a review.

christeefy:
I may have found a bug with ONNX's LRN reference implementation (see https://github.com/onnx/onnx/issues/7805). While the associated PR is being reviewed, I've regenerated the expected values with a fixed version in my local `onnx` python package. With that, the expected values matches those generated by `burn` codegen and the integration tests now pass.

We can hold off this PR merge until the changes in the `onnx` repo are merged. Lmk what you think.

antimora (approved):
All feedback addressed. Implementation is correct per the ONNX spec, tests are thorough (12 unit, 2 snapshot, 3 integration), and the BF16/rank/size validation logic is clean.

The ONNX reference evaluator bug (onnx/onnx#7805) doesn't affect us since our codegen follows the spec directly. Good to merge.

antimora:
Minor follow-up (non-blocking): consider adding a comment in `lrn.py` noting that the expected values were generated with a patched `onnx` package due to onnx/onnx#7805. That way future contributors won't get confused if re-running the script with stock `onnx==1.19.0` produces different outputs. Can be removed once the upstream fix is released.

PR #294, opened 2026-04-06
Implement NonMaxSuppression operator

### LINKED ISSUE #164: Implement NonMaxSuppression operator
## Operator

**NonMaxSuppression** - filters bounding boxes by IoU threshold. Essential for object detection models.

- [ONNX spec](https://onnx.ai/onnx/operators/onnx__NonMaxSuppression.html)
- Currently listed as `unsupported::NonMaxSuppressionNode` in `node.rs`

## Inputs
- `boxes` (T) - bounding boxes [batch, num_boxes, 4]
- `scores` (T) - scores [batch, num_classes, num_boxes]
- `max_output_boxes_per_class` (T1, optional)
- `iou_threshold` (T, optional)
- `score_threshold` (T, optional)

## Attributes
- `center_point_box` (int, default 0) - box format (0=corner, 1=center)

## Models unblocked
- YOLO variants
- SSD
- Faster R-CNN
- Any object detection model with post-processing baked into ONNX

jcwal1516:
### Checklist

- [ ] Confirmed that `cargo xtask validate` command has been executed successfully. Attempted locally, but it currently fails on an unrelated existing typo in `crates/onnx-ir/src/node/einsum.rs` also present on `upstream/main`.
- [x] For new/modified ONNX operators: verified implementation matches `onnx-spec/ops/NonMaxSuppression.md`.

### Related Issues/PRs

- Closes #164

### Changes

This PR implements `NonMaxSuppression` end to end across the import pipeline.

- add the `onnx-ir` node, config extraction, input validation, and output shape inference
- register the operator in the IR node enum and processor registry, removing it from unsupported nodes
- add `burn-onnx` codegen for greedy NMS with host-driven control flow and tensor-based IoU computation
- support both `center_point_box=0` and `center_point_box=1` box formats
- handle omitted optional inputs for `max_output_boxes_per_class`, `iou_threshold`, and `score_threshold`
- update `SUPPORTED-ONNX-OPS.md` to mark the operator as supported

### Testing

Verified locally with:

- `cargo test -p onnx-ir non_max_suppression -- --nocapture`
- `cargo test -p burn-onnx non_max_suppression -- --nocapture`
- `cargo test -p onnx-tests non_max_suppression -- --nocapture`
- `cargo fmt --check`
- `cargo check -p onnx-ir -p burn-onnx -p onnx-tests --tests`

Additional note:

- `cargo xtask validate` was run, but it fails on an unrelated existing `typos` check in `crates/onnx-ir/src/node/einsum.rs:990` (`"ij,jk->iz"`), which is already present on `upstream/main`.
- `./.venv/bin/python` is not present in this checkout, so validation used the workspace Rust/Cargo toolchain.


antimora:
@jcwal1516, thanks for adding a new op. 

For this PR, I think the primitive Non Max Suppression operator is missing from Burn. Looking at how much code would be generated, I am hesitant to support  in this form. We should add this op into Burn and use it in the generated code. You can submit a PR in burn or create an issue. 

For a next missing op, I would encourage to work on a simpler op that perhaps has some equivalent burn op (or at least with a few primitive ops).

jcwal1516:
Thanks for the feedback. I missed it earlier but looked into this further and found that burn-vision has an NMS primitive (Nms trait on Tensor<B, 2>). 

I revised the codegen to call burn-vision's NMS per batch/class instead of generating the full greedy algorithm inline. The generated code went from ~230 lines to ~80 lines (most of which is box format conversion).

One thing I noticed is burn's feature wiring doesn't propagate backend features to burn-vision. Enabling burn/vision + burn/ndarray doesn't enable burn-vision/ndarray, so the VisionBackend impl for NdArray isn't available. I worked around it by depending on burn-vision directly in onnx-tests, but this might be worth a small upstream fix so that burn = { features = ["vision", "ndarray"] } just works.

The main trade-off is models containing NonMaxSuppression now require B: VisionBackend instead of B: Backend, and downstream crates need burn with the vision feature enabled. I added infrastructure for nodes to declare extra trait bounds so this is handled automatically in the generated code.

Let me know if this approach works for the project or if you'd prefer a different direction, I am happy to adjust.

antimora:
@jcwal1516 thank you for addressing my initial concern. 

Yeah it's strange that "Enabling burn/vision + burn/ndarray doesn't enable burn-vision/ndarray"; we definitely need to file or fix upstream; we can't merge without it. 

I am CCing @mertalev (author) and @laggui on this.

laggui:
> Yeah it's strange that "Enabling burn/vision + burn/ndarray doesn't enable burn-vision/ndarray"; we definitely need to file or fix upstream; we can't merge without it.

Looks like none of the backend features are propagated to `burn-vision`. We should fix it. Right now the vision module is included as part of `burn-core` but I think we'll have to move it up to the `burn` crate only for that.

antimora:
> > Yeah it's strange that "Enabling burn/vision + burn/ndarray doesn't enable burn-vision/ndarray"; we definitely need to file or fix upstream; we can't merge without it.
> 
> Looks like none of the backend features are propagated to `burn-vision`. We should fix it. Right now the vision module is included as part of `burn-core` but I think we'll have to move it up to the `burn` crate only for that.

I filed an issue in Burn: https://github.com/tracel-ai/burn/issues/4742

antimora:
I am going to pull the latest burn rev and I'll let you know rebase

antimora:
The latest burn rev is updated. Please rebase and resolve the conflicts.

antimora:
I am waiting for @laggui complete a backend refactor that will affect backend extensions

jcwal1516:
Sorry, I completely forgot about this until I picked up an old project again. I use YOLO models for some pathology work, so having NonMaxSuppression work in burn-onnx would be pretty useful for end-to-end ONNX exports.

I saw the last update here was waiting on the backend refactor. That has since landed in #403, so I updated the implementation locally against current main. The focused and official ONNX tests are passing.

The current Burn pin still doesn’t include the latest burn-vision feature forwarding for WGPU/Metal, although that has been fixed upstream now. I see #422 is currently working on the Burn bump.

Should I update this PR or open a fresh one? Also, should I wait for the Burn bump before pushing, or rebase the NMS work now and keep the dependency change separate? Let me know happy for either!

antimora:
Lets reopen a new one since this has diverged a lot.

Issue #160, opened 2026-02-12
Implement quantization operators (DequantizeLinear, QuantizeLinear, QLinearConv, QLinearMatMul, DynamicQuantizeLinear)


antimora:
## Context

Gap analysis against reveals quantization as the biggest strategic gap. tract has a full INT8 inference pipeline. burn-onnx only has MatMulInteger.

Nearly all production edge deployments use INT8 quantization. This is the primary blocker for production model deployment.

## Operators needed

| Operator | ONNX Spec | Status |
|---|---|---|
| **DequantizeLinear** | [spec](https://onnx.ai/onnx/operators/onnx__DequantizeLinear.html) | `unsupported::DequantizeLinearNode` |
| **QuantizeLinear** | [spec](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html) | `unsupported::QuantizeLinearNode` |
| **DynamicQuantizeLinear** | [spec](https://onnx.ai/onnx/operators/onnx__DynamicQuantizeLinear.html) | `unsupported::DynamicQuantizeLinearNode` |
| **QLinearConv** | [spec](https://onnx.ai/onnx/operators/onnx__QLinearConv.html) | `unsupported::QLinearConvNode` |
| **QLinearMatMul** | [spec](https://onnx.ai/onnx/operators/onnx__QLinearMatMul.html) | `unsupported::QLinearMatMulNode` |
| **ConvInteger** | [spec](https://onnx.ai/onnx/operators/onnx__ConvInteger.html) | `unsupported::ConvIntegerNode` |

## Suggested implementation order

1. **DequantizeLinear** + **QuantizeLinear** - foundation for all quantized models
2. **QLinearMatMul** - quantized matrix multiplication
3. **QLinearConv** - quantized convolution
4. **DynamicQuantizeLinear** - runtime quantization
5. **ConvInteger** - integer convolution

## Models unblocked
- Any INT8/quantized ONNX model
- tract's inceptionv1_quant test case
- Production-optimized models from ONNX Runtime quantization toolkit

## Notes
This depends on Burn backend support for quantized tensor operations. May need coordination with the upstream Burn framework.

christeefy:
Hey @antimora, I'm interested in implementing the rest of these ops, and support more quantized models 🙂

[...]

PR #333, opened 2026-04-13
Implement `QLinearMatMul` operator

christeefy:
### Checklist

### Related Issues/PRs

#160

### Changes

Implements the `QLinearMatMul` operator. 

### Testing

- Unit testing in `onnx-ir` to validate as much input invariants as possible.
- Unit testing in the `burn-onnx` crate, accounting for cases where the scale and zero_point inputs come in various ranks, relative to their corresponding tensor's rank. 
- Integration testing in the `onnx-tests` crate. 
   - Four ONNX models and tests are generated:
      1. Scalar scale and zero_points
      2. Vector scale and zero_points
      3. N-D (broadcastable) scale and zero_points
      4. Saturating case ([bug](https://github.com/onnx/onnx/issues/7835) found in ONNX reference implementation once again)

### Notes and Limitations: 
- The codegen panics when dealing with QFloats, as they are currently [not supported](https://github.com/christeefy/burn-onnx/blob/a0ee4510a9d4188ed1ca3015bdbb28ecd1f416e3/crates/burn-onnx/src/burn/argument_helpers.rs#L22).
- The [spec](https://onnx.ai/onnx/operators/onnx__QLinearMatMul.html) also specifies the use of unsigned zero float8 variants (`float8e5m2fnuz` and `float8e4m3fnuz`). These types are not checked for as they are not yet modelled within `burn`. 

The above issues have been documented as comments or xfail-ed unit tests.


crates/onnx-ir/src/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,470 @@
...
+    #[test]
+    fn test_scale_rank_exceeds_tensor_rank() {
+        let mut node = build_base_node();
+        // Scale rank exceeds tensor rank
+        node.inputs[0].ty = ArgType::Tensor(TensorType::new(DType::I8, 2, None)); // a rank 2
+        node.inputs[1].ty = ArgType::Tensor(TensorType::new(DType::F32, 3, None)); // a_scale rank 3
+        node.inputs[2].ty = ArgType::Tensor(TensorType::new(DType::F32, 3, None)); // a_zero_point rank 3
+
+        let result = QLinearMatMulProcessor.infer_types(&mut node, 21, &OutputPreferences::new());
+        assert!(matches!(result, Err(ProcessError::TypeMismatch { .. })));
```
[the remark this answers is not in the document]
christeefy: Fixed. This test has been renamed to `test_*_scale_zp_and_tensor_rank_mismatch`.

crates/onnx-ir/src/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,470 @@
...
+            .input_scalar_f32("a") // Wrong: should be tensor
+            .input_tensor_f32("a_scale", 0, None)
+            .input_tensor_i32("a_zero_point", 0, None)
+            .input_tensor_i32("b", 2, None)
+            .input_tensor_f32("b_scale", 0, None)
+            .input_tensor_i32("b_zero_point", 0, None)
+            .input_tensor_f32("y_scale", 0, None)
+            .input_tensor_i32("y_zero_point", 0, None)
+            .output_tensor_f32("y", 2, None)
+            .build();
```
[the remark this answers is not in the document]
christeefy: Fixed as well. Tests that expect an error now also match on keywords to ensure the proper error path is being tested.

crates/onnx-ir/src/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,470 @@
...
+
+        // Validate ranks for scale and zero point against corresponding input tensor
+        for (tensor, scale, name) in [(a, a_scale, "a"), (b, b_scale, "b")] {
+            if scale.ty.rank() > tensor.ty.rank() {
+                return Err(ProcessError::TypeMismatch {
+                    expected: "The ranks for the scale and zero point must be less than or equal to its corresponding tensor".to_string(),
+                    actual: format!("{name} has rank {} while {name}_scale has rank {}", tensor.ty.rank(), scale.ty.rank()),
+                });
+            }
+        }
```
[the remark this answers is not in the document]
christeefy: Validation logic has been made more specific to capture various cases instead of simply ensuring the scale and zero point ranks do not exceed their corresponding tensor's rank.

crates/onnx-ir/src/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,470 @@
...
+            (y_zero_point, "y_zero_point"),
+        ] {
+            let dtype = input.ty.elem_type();
+            match dtype {
+                DType::I8 | DType::U8 => {}
+                // FIXME: Support unsized F8 types (F85ME2UZ and F8M5E2UZ) as required by the spec
+                DType::QFloat(quant_scheme)
+                    if opset >= 21
+                        && matches!(quant_scheme.value, QuantValue::E5M2 | QuantValue::E4M3) => {}
+                _ => {
```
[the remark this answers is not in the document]
christeefy: Shifted `ProcessError::TypeMismatch` to the `DType::QFloat(...)` branch, with a `FIXME` for future work.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        quote! {
+            // Dequantize inputs
+            #reshape_a_scale_and_zp
+            #reshape_b_scale_and_zp
+            let a_dequantized = #a_scale * (#a_float - #a_zero_point_float);
+            let b_dequantized = #b_scale * (#b_float - #b_zero_point_float);
+
+            let output_tensor = a_dequantized.matmul(b_dequantized);
+
```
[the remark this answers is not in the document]
christeefy: Reused codegen logic from `crate/burn-onnx/.../matmul.rs`, and verified with an additional snapshot test.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,357 @@
...
+        let a_arg = self.inputs.first().unwrap();
+        let a_scale_arg = self.inputs.get(1).unwrap();
+        let a_zero_point_arg = self.inputs.get(2).unwrap();
+        let b_arg = self.inputs.get(3).unwrap();
+        let b_scale_arg = self.inputs.get(4).unwrap();
+        let b_zero_point_arg = self.inputs.get(5).unwrap();
+        let y_scale_arg = self.inputs.get(6).unwrap();
+        let y_zero_point_arg = self.inputs.get(7).unwrap();
+
+        let output = arg_to_ident(self.outputs.first().unwrap());
```
antimora, line 30: Eight `.unwrap()`s in library codegen violate the no-unwrap rule. `InputSpec::Exact(8)` protects you through the processor, but `QLinearMatMulNodeBuilder` can build the node directly (the snapshot tests do this). Prefer direct indexing or a destructure:

```rust
let [a_arg, a_scale_arg, a_zero_point_arg, b_arg, b_scale_arg, b_zero_point_arg, y_scale_arg, y_zero_point_arg] =
    self.inputs.as_slice()
    else { unreachable!("QLinearMatMul requires 8 inputs") };
```

Or, even better, add named accessors to `QLinearMatMulNode` so all 16 positional call sites (here and in `reshape_scale_and_zp`/`to_float` args) become `self.a()`, `self.a_scale()`, etc. That also kills the risk of silently misinterpreting inputs if the order ever drifts.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,357 @@
...
+            #reshape_b_scale_and_zp
+            let a_dequantized = #a_scale * (#a_float - #a_zero_point_float);
+            let b_dequantized = #b_scale * (#b_float - #b_zero_point_float);
+
+            #matmul_statement
+
+            // Quantize output
+            #reshape_y_scale_and_zp
+            let #output = (output_tensor / #y_scale).round();
+            let #output = (#output + #y_zero_point_float)#clamp_expr.int().cast(#output_dtype);
```
antimora, line 92: Bare `.int()` here violates the explicit-dtype rule. Even though `.cast(#output_dtype)` follows, `.int()` materializes a tensor with the backend's default `IntElem`, which varies across backends (CPU vs GPU). The subsequent `cast` from that default-int to U8/I8 can have backend-dependent saturation semantics.

Suggested: stay in float, clamp, cast straight to the target dtype:

```rust
let #output = (#output + #y_zero_point_float)#clamp_expr.cast(#output_dtype);
```

Floats round cleanly to integers via cast, and the clamp already bounds the range before conversion.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,357 @@
...
+        let a_float = to_float(a_arg, a);
+        let a_zero_point_float = to_float(a_zero_point_arg, a_zero_point);
+        let b_float = to_float(b_arg, b);
+        let b_zero_point_float = to_float(b_zero_point_arg, b_zero_point);
+        let y_zero_point_float = to_float(y_zero_point_arg, y_zero_point);
+
+        let mut output_rank = a_arg.ty.rank().max(b_arg.ty.rank());
+        if a_arg.ty.rank() == 1 || b_arg.ty.rank() == 1 {
+            output_rank -= 1;
+        }
```
antimora: This re-derives logic already in `matmul_output_rank` (which you just generalized in 63ae1912). Two options to avoid the second source of truth:

1. Re-export `matmul_output_rank` as `pub` from `onnx-ir::node::matmul` and call it here.
2. Just read it from the already-inferred output: `self.outputs[0].ty.rank()` — `infer_types` has already stored the correct value.

Option 2 is zero-touch.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,357 @@
...
+fn to_float(arg: &Argument, token_stream: TokenStream) -> TokenStream {
+    let needs_cast = arg.ty.elem_type().is_int() || arg.ty.elem_type().is_uint();
+    if !needs_cast {
+        token_stream
+    } else if arg.ty.is_scalar() {
+        quote! { (#token_stream as f32) }
+    } else {
+        quote! { #token_stream.float() }
+    }
+}
```
antimora, line 121: Silent fallthrough: if a dtype we didn't anticipate reaches this function, we emit `#token_stream` unmodified and the generated Rust fails to compile with no hint that `to_float` was the source.

More concretely: opset-21 scales can be BF16 or F16 (`infer_types` accepts them at line 133), but this path always casts operands to F32 and emits scales as-is. `a_scale * (a_float - a_zero_point_float)` then mixes F32 with BF16/F16 and either fails to compile or silently upcasts depending on the backend. The integration tests all use F32 scales so this is uncovered.

Either (a) restrict scale dtype to F32 in the processor until codegen handles half-precision, or (b) cast scales to F32 here. Also consider an `unreachable!` in the fallthrough to make the contract visible.

crates/onnx-ir/src/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,481 @@
...
+        // Validate dtypes for zero points inputs
+        for (input, name) in [
+            (a_zero_point, "a_zero_point"),
+            (b_zero_point, "b_zero_point"),
+            (y_zero_point, "y_zero_point"),
+        ] {
+            let dtype = input.ty.elem_type();
+            match dtype {
+                DType::I8 | DType::U8 => {}
+                // FIXME: Support unsized F8 types (F85ME2UZ and F8M5E2UZ) as required by the spec
```
antimora: Typos in the type names will make this FIXME ungreppable. The ONNX spec calls them `FLOAT8E4M3FNUZ` and `FLOAT8E5M2FNUZ` (note `FNUZ` suffix, and E comes before the mantissa bits). Worth getting these right so a future reader can grep for the canonical names.

crates/onnx-ir/src/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,481 @@
...
+                    return Err(ProcessError::TypeMismatch {
+                        expected: "I8 or U8 operand tensor dtypes. F8 is not yet supported."
+                            .to_string(),
+                        actual: format!("{name}: {:?}", quant_scheme.value),
+                    });
+                }
+                _ => {
+                    return Err(ProcessError::TypeMismatch {
+                        expected: if opset >= 21 {
+                            "Only I8, U8, F8M5E2, F8M4E3 tensor dtypes are supported".to_string()
```
antimora: This message contradicts the branch above it. The arm at line 72-83 rejects `E5M2` / `E4M3` with a "not yet supported" error, but this fallback tells the user "Only I8, U8, F8M5E2, F8M4E3 tensor dtypes are supported" — implying F8 *is* supported. Confusing.

Also the names: `F8M5E2` and `F8M4E3` aren't ONNX-spec names (should be `FLOAT8E5M2` / `FLOAT8E4M3FN`). Suggestion: drop F8 from this message entirely and leave the dedicated F8 branch above to handle that case with its own clear "not yet implemented" wording.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
```diff
@@ -0,0 +1,357 @@
...
+        &self.outputs
+    }
+
+    fn forward(&self, scope: &mut ScopeAtPosition<'_>) -> TokenStream {
+        // ASSUMPTIONS (per the ONNX spec)
+        // 1. Scale and zero point input for a given operand have the same shape.
+
+        // NOTE: QFloat operand tensors are rejected by the onnx-ir processor
+        //       for now until they are supported in codegen.
+        //       As such, operand tensors and their zero points of type I8 or U8.
```
antimora: Sentence fragment (missing verb) and the framing is a bit off. How about:

```
// QFloat operands are rejected upstream (see onnx-ir::qlinear_matmul),
// so here both operand tensors and their zero points are guaranteed to be I8 or U8.
```

crates/onnx-tests/tests/qlinear_matmul/qlinear_matmul.py
```diff
@@ -0,0 +1,191 @@
...
+        "saturate",
+        a=np.array([[120, 120, 120, 120], [0, 0, 0, 0]], dtype=np.uint8),
+        a_scale=np.array(0.1, dtype=np.float32),
+        a_zero_point=np.array(2, dtype=np.uint8),
+        b=np.array([[120, 5], [120, 5], [120, 5], [120, 5]], dtype=np.uint8),
+        b_scale=np.array(0.2, dtype=np.float32),
+        b_zero_point=np.array(3, dtype=np.uint8),
+        y_scale=np.array(0.3, dtype=np.float32),
+        y_zero_point=np.array(4, dtype=np.uint8),
+    )
```
antimora, line 207: Small test gap: all four integration models use U8 end-to-end. The I8 path through `clamp(-128, 127)` and the I8 cast only get exercised by snapshot tests, which don't catch runtime clamp/cast bugs. Worth adding one I8 model, and one opset-10 model — the opset-gating logic in `infer_types` is unit-tested but not hit at the integration layer. Not a blocker, but cheap to add while you're here.

antimora (commented):
Nice work on this one. The processor split, rank-compat matrix, and the saturate test (with hand-computed expected values to dodge the ONNX ref bug) all look solid. A few things I'd like to see addressed before merge, mostly in the codegen. Copilot's earlier points on QFloat / output-rank / matmul rank alignment look addressed.

antimora (changes requested):

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
christeefy: Added named accessors.

crates/onnx-ir/src/node/qlinear_matmul.rs
christeefy: Updated comments.

crates/onnx-tests/tests/qlinear_matmul/qlinear_matmul.py
christeefy: Good points! Added integration tests case 5 & 6.

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
christeefy: I'm curious what the some backend-dependent edge cases may look like, given that the values are clamped to the ranges of U8/I8, and default types are likely I32 and above. 

Also, does this mean that instructions around `.cast`-ing after `.int()` or `.float()` in the agent docs is outdated?

I removed it and the integration tests were failing: 
```rs
) -> Tensor<B, 3, Int> {
   |          ----------------- expected `burn::Tensor<B, 3, burn::prelude::Int>` because of return type
...
73 |         qlinearmatmul1_out1
   |         ^^^^^^^^^^^^^^^^^^^ expected `Tensor<B, 3, Int>`, found `Tensor<B, 3>`
   |
   = note: expected struct `burn::Tensor<_, _, burn::prelude::Int>`
              found struct `burn::Tensor<_, _, burn::prelude::Float>`
```
So this suggests that `.int()` is required 🤔 

From my limited understanding of `burn` tensors, casting changes the precision, but `.int()` or `.float()` changes the "kind" of the tensor. Is that a correct mental model?

crates/burn-onnx/src/burn/node/qlinear_matmul.rs
christeefy: Thanks for catching this edge case (the input space is so wide that there's just so much to cover @.@...)

I've done the following:
- casted the `scale`s to F32
- added the `unreachable!` fallthrough
- appended `.cast(DType::F32)` after `.float()` for int-dtypes, and
- use `f32::from` instead of `as f32` since `half::{bf16, f16}` are used in the integration tests

antimora (approved):
Looks good to me; Thank you for your contribution!

PR #357, opened 2026-04-17
Adds support for the ONNX ML Scaler operator

saminiin:
Implementation matches https://onnx.ai/onnx/operators/onnx_aionnxml_Scaler.html

Formula: Y = (X - offset) * scale

SUPPORTED-ONNX-OPS.md: Scaler added


crates/burn-onnx/src/burn/node/scaler.rs
```diff
@@ -0,0 +1,325 @@
...
+                    (true, true) => {
+                        let offset_values: Vec<_> = self.config.offset.as_ref().unwrap().to_vec();
+                        let scale_values: Vec<_> = self.config.scale.as_ref().unwrap().to_vec();
+                        if offset_values.len() != scale_values.len() {
+                            panic!(
+                                "Scaler: offset and scale must have the same number of elements, got offset={}, scale={}",
+                                offset_values.len(),
+                                scale_values.len()
+                            );
+                        }
```
antimora: Disagree with Copilot's suggestion to keep this panic. Length mismatch is already rejected by `ScalerProcessor::extract_config` (onnx-ir/src/node/scaler.rs:113-124) before codegen ever runs, so this branch is unreachable. A `panic!` here would only fire if IR validation regressed, and it would crash the user's build with no file/line context. Prefer `unreachable!("validated in onnx-ir")` or just delete the check.

crates/burn-onnx/src/burn/node/scaler.rs
```diff
@@ -0,0 +1,325 @@
...
+                                #input_expr * scale_tensor
+                            }
+                        }
+                    }
+                    (false, false) => {
+                        quote! { #input_expr }
+                    }
+                }
+            }
+            ty => panic!("Scaler: unexpected input type {ty:?}"),
```
antimora: Same story as the length-mismatch panic above: `infer_types` already returns `ProcessError::TypeMismatch` for non-tensor inputs, so this arm is unreachable. Swap for `unreachable!(...)` or remove and let the match be exhaustive.

crates/burn-onnx/src/burn/node/scaler.rs
```diff
@@ -0,0 +1,325 @@
...
+                    (0..input_rank.saturating_sub(1))
+                        .map(|_| quote! { 1usize })
+                        .chain(std::iter::once(quote! { #num_features }))
+                        .collect()
+                };
+
+                match (has_offset, has_scale) {
+                    (true, true) => {
+                        let offset_values: Vec<_> = self.config.offset.as_ref().unwrap().to_vec();
+                        let scale_values: Vec<_> = self.config.scale.as_ref().unwrap().to_vec();
```
antimora: Minor: these `.unwrap()`s are logically safe (guarded by `has_offset`/`has_scale` two lines up), but restructuring as `match (&self.config.offset, &self.config.scale)` over `Option` directly would remove the unwraps and make the guarantee structural. Same on lines 73 and 89.

crates/onnx-ir/src/node/scaler.rs
```diff
@@ -0,0 +1,166 @@
...
+            match key.as_str() {
+                "scale" => {
+                    if let AttributeValue::Float32s(floats) = value {
+                        scale = Some(floats.clone());
+                    }
+                }
+                "offset" => {
+                    if let AttributeValue::Float32s(floats) = value {
+                        offset = Some(floats.clone());
+                    }
```
antimora, line 117: Silent fallback: if `scale` or `offset` is present but the `AttributeValue` variant isn't `Float32s` (e.g. a malformed model encodes a single `Float32`, or an `Int64s`), the `if let` falls through and the attribute is treated as absent. The user sees "no scale configured" behavior rather than a clear error. Consider an `else` arm that returns `ProcessError::InvalidAttribute { name: "scale", reason: "expected Float32s" }`.

crates/onnx-ir/src/node/scaler.rs
```diff
@@ -0,0 +1,166 @@
...
+            });
+        }
+
+        Ok(ScalerConfig::new(scale, offset))
+    }
+
+    fn build_node(&self, builder: RawNode, opset: usize) -> Node {
+        let config = self
+            .extract_config(&builder, opset)
+            .expect("ScalerProcessor: config extraction failed");
```
antimora, line 144: `build_node` turns any `ProcessError` from `extract_config` into a panic via `.expect`. In practice the pipeline runs `extract_config` earlier and surfaces the `Result`, so this second call is effectively infallible — but that's an invariant of the pipeline, not of this function. Worth a short comment noting why the `.expect` is safe, or restructuring so `build_node` takes the already-validated config.

crates/onnx-ir/src/node/scaler.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn test_scaler_node_builder() {
+        let config = ScalerConfig::new(Some(vec![2.0]), Some(vec![1.0]));
+        let node = ScalerNode::new("test_scaler".to_string(), vec![], vec![], config);
+
+        assert_eq!(node.name, "test_scaler");
+        assert_eq!(node.inputs.len(), 0);
+        assert_eq!(node.outputs.len(), 0);
+        assert!(node.config.scale.is_some());
+        assert!(node.config.offset.is_some());
+    }
```
antimora, line 206: These two tests only exercise `ScalerConfig::new` and the derived builder — they never go through `ScalerProcessor::extract_config` or `infer_types`. The length-mismatch `ProcessError` path (L113-124) and the dtype rejection in `infer_types` (L67-74) have no unit coverage, so a regression in either would pass CI. Consider adding tests that build a `RawNode` with attributes and call the processor methods directly.

crates/onnx-tests/tests/scaler/scaler.py
```diff
@@ -0,0 +1,72 @@
...
+    print("\nInput:")
+    print(input_data)
+    print("\nOutput (Y = (X - 1.0) * 2.0):")
+    print(result[0])
+    print("\nExpected:")
+    print((input_data - 1.0) * 2.0)
+
+
+if __name__ == "__main__":
+    main()
```
antimora, line 132: Integration coverage is thin — only 2D F32 with length-1 `scale`/`offset`. The interesting paths in the codegen are (a) per-feature vectors (`scale.len() > 1`, which exercises the `[1, ..., 1, F]` reshape-to-last-axis broadcast) and (b) integer / F64 input dtypes (which exercise the `.cast(DType::F32)` path). Snapshots verify those compile; only an integration test verifies the arithmetic is right. Worth adding at least a rank-3 per-feature case and an I64 case.

antimora (commented):
Reviewed the Scaler addition. Most of Copilot's points are already addressed in the latest commit (static_shape preservation, input dtype validation in `infer_types`, scale/offset length check in `extract_config`, spec URL). I disagree with Copilot's suggestion to add a length-mismatch `panic!` inside codegen — see inline. A few additional items below, mostly about silent fallbacks and test coverage.

Nothing blocking; the op-level logic and registration look correct.

antimora (changes requested):

crates/burn-onnx/src/burn/node/scaler.rs
saminiin: Yes, that's true. I'll remove that futile check.

crates/onnx-ir/src/node/scaler.rs
saminiin: You're correct. I'll fix it.

crates/onnx-tests/tests/scaler/scaler.py
saminiin: I'll add more tests

antimora (approved):
LGTM; thank you for addressing my feedback

saminiin:
Thanks for good and valuable comments!

PR #370, opened 2026-04-24
Bump all dependencies to latest

antimora:
## Summary

Two fixes, both of which are needed to unblock the code-quality CI job on #369:

1. **Audit ignore for RUSTSEC-2026-0110** (`bare-metal` unmaintained). Transitive via `cortex-m 0.7` in the `raspberry-pi-pico` example. `cortex-m` has not yet bumped to `bare-metal 1.0`, and the advisory is informational (unmaintained), not a vulnerability. Without this, `cargo audit` fails every PR.
2. **Workspace dep bumps** picked up incidentally:
   - `toml` 1.1.0 -> 1.1.2
   - `tracel-xtask` =4.15.1 -> =4.17.2 (pinned)
   - `wasm-bindgen` 0.2.114 -> 0.2.118 (image-classification-web example)
   - `Cargo.lock` refresh (~30 transitive bumps)

Burn git rev (`37e7e613286b1892480e5cc0979ae221685733f9`) is already at the latest main head.

## Test plan
- [x] `cargo xtask check audit` passes locally
- [x] `cargo test -p onnx-ir -p burn-onnx -p onnx-tests -p onnx-official-tests` - all 21 test runs pass, 0 failures
- [x] `cargo build -p xtask` succeeds with the new `tracel-xtask 4.17.2`
- [x] `cargo fmt --check` passes on modified crates

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -13,6 +13,7 @@ ignore = [
     "RUSTSEC-2026-0097", # `rustls-webpki` advisory (transitive via ureq/tch).
     "RUSTSEC-2026-0098", # `rustls-webpki` advisory (transitive via ureq/tch).
     "RUSTSEC-2026-0099", # `rustls-webpki` advisory (transitive via ureq/tch).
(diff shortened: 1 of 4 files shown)


antimora:
Folded into #369 so the feature PR's CI can pass on its own.

PR #376, opened 2026-04-27
Fix burn book link in readme

BenFradet:
### Checklist


antimora:
Hi @BenFradet,

We haven't published the updated book yet. We changed burn-import to burn-onnx.

Here is the book section in the current form for your reference: https://github.com/tracel-ai/burn/blob/main/burn-book/src/onnx-import.md

PR #378, opened 2026-04-28
Add SVMRegressor operator

saminiin:
-Confirmed that cargo xtask validate command has been executed and fixes made.
-Confirm relevant documentation has been updated.
-Implementation matches onnx-spec/ops/SVMRegressor.md.

Adds support for the SVMRegressor ONNX operator (opset 1). Implements all five kernels: LINEAR, RBF, POLY, SIGMOID, and LOGISTIC, plus the SOFTMAX and SOFTMAX_ZERO post-transforms. PROBIT emits a compile_error! since it has no Burn equivalent. 

Seven codegen snapshot tests cover each kernel and post-transform variant. Two integration tests run the generated model against known-good outputs produced by onnx.reference.ReferenceEvaluator: one for the LINEAR kernel and one for RBF.


crates/onnx-ir/src/node/svmregressor.rs
```diff
@@ -0,0 +1,211 @@
...
+/// Performs regression using Support Vector Machine (SVM) with various kernel types.
+#[allow(clippy::too_many_arguments)]
+#[derive(Debug, Clone, Default, new)]
+pub struct SVMRegressorConfig {
+    /// Coefficients for the support vector in the decision function.
+    pub coefficients: Option<Vec<f32>>,
+    /// Parameters for the kernel function (gamma, coef0, degree).
+    pub kernel_params: Option<Vec<f32>>,
+    /// Type of kernel function: LINEAR, POLY, RBF, SIGMOID. Default: `"LINEAR"`.
+    pub kernel_type: Option<String>,
```
antimora: Should be enum

crates/onnx-ir/src/node/svmregressor.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Number of support vectors. Default: `0`.
+    pub n_supports: Option<i64>,
+    /// Flag indicating one-class SVM anomaly detection mode. Default: `0` (disabled).
+    pub one_class: Option<i64>,
+    /// How to transform the output: NONE, SOFTMAX, LOGISTIC, SOFTMAX_ZERO, PROBIT. Default: `"NONE"`.
+    pub post_transform: Option<String>,
```
antimora: Should be enum

crates/burn-onnx/src/burn/node/svmregressor.rs
```diff
@@ -0,0 +1,418 @@
...
+                let [batch, _] = y.dims();
+                let zero_col = burn::tensor::Tensor::<B, 2>::zeros(
+                    burn::tensor::Shape::new([batch, 1usize]),
+                    &y.device(),
+                );
+                let combined = burn::tensor::Tensor::cat(alloc::vec![y, zero_col], 1);
+                let e = combined.exp();
+                e.clone().narrow(1, 0, 1) / e.sum_dim(1)
+            } },
+            "PROBIT" => quote! {
```
antimora: Emitting `compile_error!()` from generated code means `onnx2burn` succeeds, then the user sees a confusing rust compile error far from the original ONNX node. The `one_class=1` rejection at the top of `extract_config` is the right pattern, same here for `post_transform == "PROBIT"`.

crates/burn-onnx/src/burn/node/svmregressor.rs
```diff
@@ -0,0 +1,418 @@
...
+                _ => quote! { #cast_expr },
+            }
+        };
+
+        // Apply post-transform if needed
+        let function = match post_transform {
+            "NONE" => kernel_computation,
+            "LOGISTIC" => quote! { { let y = #kernel_computation; (y.neg().exp() + 1.0).recip() } },
+            "SOFTMAX" => {
+                quote! { { let y = #kernel_computation; let e = y.exp(); e.clone() / e.sum_dim(1) } }
```
antimora: Since coefficients get reshaped to `[n_supports, 1]`, the kernel output is `[N, 1]`. softmax over dim=1 of `[N, 1]` is just ones, always. So this branch is silently broken for the only shape we ever produce. Either reject `post_transform == "SOFTMAX"` for the single-target case in `extract_config`, or fold it in once multi-target lands.

crates/burn-onnx/src/burn/node/svmregressor.rs
```diff
@@ -0,0 +1,418 @@
...
+                    }
+                }
+                "RBF" => {
+                    let gamma = self
+                        .config
+                        .kernel_params
+                        .as_ref()
+                        .and_then(|p| p.first())
+                        .copied()
+                        .unwrap_or(0.1);
```
antimora: `unwrap_or(0.1)` is arbitrary. The ONNX spec doesn't define a default for `kernel_params`, and scikit/ORT use `1/n_features`. If `kernel_params` is missing for a kernel that needs it, the model is effectively malformed. Cleaner to reject in `extract_config` than silently substitute a guess. Same applies to POLY (line 135) and SIGMOID (line 158).

antimora (commented):
A few things on top of what Copilot already flagged.

crates/onnx-tests/tests/svmregressor/mod.rs
```diff
@@ -0,0 +1,62 @@
+// Import the shared macro
+use crate::include_models;
+include_models!(svmregressor, svmregressor_rbf);
```
antimora: Integration tests only hit LINEAR and RBF. POLY, SIGMOID, and the post-transforms have only snapshot tests, which check the generated rust looks right but not that it computes correct values. Given the SOFTMAX-on-`[N, 1]` issue I left on the codegen file, at least one post-transform integration test would have caught it.

crates/onnx-ir/src/node/svmregressor.rs
```diff
@@ -0,0 +1,211 @@
...
+        _opset: usize,
+        _output_preferences: &OutputPreferences,
+    ) -> Result<(), ProcessError> {
+        // Per the ONNX spec, output Y is always tensor(float) regardless of input type.
+        // Input T may be tensor(double), tensor(float), tensor(int32), or tensor(int64).
+        // Output shape is [N] for [N, C] input (one score per sample).
+        node.outputs[0].ty = ArgType::Tensor(TensorType {
+            dtype: DType::F32,
+            rank: 1,
+            static_shape: None,
```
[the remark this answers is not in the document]
antimora: +1, this one's a real foot-gun. Agreed on rejecting rank-1 with a clear error rather than silently unsqueezing. The `[C]` form is rare and quietly reshaping the input would be surprising. Allowed dtype set looks right too.

crates/onnx-ir/src/node/svmregressor.rs
```diff
@@ -0,0 +1,211 @@
...
+                "rho" => {
+                    if let AttributeValue::Float32s(floats) = value {
+                        rho = Some(floats.clone());
+                    }
+                }
+                "support_vectors" => {
+                    if let AttributeValue::Float32s(floats) = value {
+                        support_vectors = Some(floats.clone());
+                    }
+                }
```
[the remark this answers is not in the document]
antimora: Agree. The current `if let` arms quietly swallow malformed protos. `InvalidAttribute` matches what other processors do.

crates/onnx-ir/src/node/svmregressor.rs
```diff
@@ -0,0 +1,211 @@
...
+        Ok(SVMRegressorConfig::new(
+            coefficients,
+            kernel_params,
+            kernel_type,
+            n_supports,
+            one_class,
+            post_transform,
+            rho,
+            support_vectors,
+        ))
```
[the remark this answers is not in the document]
antimora: Agreed. Tying this to the multi-target case: `coefficients.len() == n_supports * n_targets` is exactly where the current codegen breaks. It hardcodes `n_targets = 1` (the `[n_supports, 1]` reshape, the final `squeeze_dim::<1>(1)`, and the rank-1 output in `infer_types`). Easiest path: validate `coefficients.len() == n_supports` here and reject multi-target for now with a clear error. Threading `n_targets` through codegen can come later.

crates/burn-onnx/src/burn/node/svmregressor.rs
```diff
@@ -0,0 +1,418 @@
...
+            .config
+            .rho
+            .as_ref()
+            .and_then(|r| r.first())
+            .copied()
+            .unwrap_or(0.0);
+        let n_supports = self.config.n_supports.unwrap_or(0) as usize;
+
+        // Reference the stored tensors (tuple access .0 and .1)
+        let field_name = Ident::new(&self.name, Span::call_site());
```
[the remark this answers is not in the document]
antimora: Good catch, I'd missed that the `field()` and `forward()` conditions aren't aligned. Agreed the cleaner fix is to reject in `extract_config` so codegen never sees the degenerate case.

crates/burn-onnx/src/burn/node/svmregressor.rs
```diff
@@ -0,0 +1,418 @@
...
+            quote! { (Tensor<B, 2>, Tensor<B, 2>) },
+            quote! {
+                let #name = (
+                    Tensor::<B, 2>::from_data(
+                        burn::tensor::TensorData::new(alloc::vec![#(#coef_data),*], [#n_supports, 1]),
+                        (device, burn::tensor::DType::F32),
+                    ),
+                    Tensor::<B, 2>::from_data(
+                        burn::tensor::TensorData::new(alloc::vec![#(#sv_data),*], [#n_supports, #n_features]),
+                        (device, burn::tensor::DType::F32),
```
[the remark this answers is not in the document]
antimora: Yep. And `n_features` is computed at codegen-time but never stored. Moving the divisibility check into `extract_config` would also let us stash `n_features` (and `n_targets`) on the config itself, so codegen isn't recomputing layout.

antimora (changes requested):

crates/burn-onnx/src/burn/node/svmregressor.rs
saminiin: Good catch, i'll fix it.

crates/onnx-tests/tests/svmregressor/mod.rs
saminiin: I'll add more tests

crates/onnx-ir/src/node/svmregressor.rs
saminiin: Fixed.

crates/onnx-ir/src/node/svmregressor.rs
saminiin: Fixed-

antimora:
@saminiin Thanks for the latest round! We picked up two outstanding items locally so we can get this into the upcoming pre-release. Summary of what changed beyond `566e39f3`:

### 1. Reject `SOFTMAX` post_transform in `extract_config` (resolves [#discussion_r3157205637](https://github.com/tracel-ai/burn-onnx/pull/378#discussion_r3157205637))

The current codegen produces `[N, 1]` (single-target), and softmax over a size-1 dim is always `1.0`. Rather than emit silently-broken predictions, we mirror the PROBIT pattern and reject in onnx-ir with a clear message pointing users at `NONE` / `LOGISTIC` / `SOFTMAX_ZERO`. The `Softmax` codegen arm becomes `unreachable!()` and its snapshot test is removed. `SOFTMAX_ZERO` stays — it's mathematically meaningful for single-target since it appends an implicit zero class.

- `crates/onnx-ir/src/node/svmregressor.rs` — new rejection block alongside PROBIT
- `crates/burn-onnx/src/burn/node/svmregressor.rs` — `Softmax` arm → `unreachable!`, dropped `test_svm_linear_softmax`

### 2. Add `SOFTMAX_ZERO` integration test (follows up [#discussion_r3157205644](https://github.com/tracel-ai/burn-onnx/pull/378#discussion_r3157205644))

LOGISTIC plus the four kernel integration tests you added cover the runtime side well. The remaining gap was a runtime check on `SOFTMAX_ZERO`. Added a sixth model and test using the same fixture; for single-target, `SOFTMAX_ZERO` equals `LOGISTIC` (`exp(y)/(exp(y)+1) = sigmoid(y)`), which the values confirm.

- `tests/svmregressor/svmregressor.py` — generates `svmregressor_softmax_zero.onnx`
- `tests/svmregressor/svmregressor_softmax_zero.onnx` — new fixture
- `tests/svmregressor/mod.rs` — `svmregressor_softmax_zero` test
- `build.rs` — registers the new fixture

All 6 integration + 6 codegen snapshots + 2 onnx-ir unit tests pass; `cargo fmt --check` and `cargo clippy -p onnx-ir -p burn-onnx` are clean.

antimora (approved):

PR #401, opened 2026-05-12
fix: accept rank-1 input in Flatten per ONNX spec

### LINKED ISSUE #349: Flatten: accept rank-1 input (14 tests)
## Summary

`Flatten` rejects 1-D input. **14 tests** `skip-codegen`.

## Error

```
Node 'flatten2' (Flatten): Flatten: input tensor must have at least 2 dimensions (got 1)
```

## Example tests

- `test_layer_normalization_2d_axis1_expanded`
- `test_layer_normalization_2d_axis1_expanded_ver18`
- `test_layer_normalization_2d_axis_negative_1_expanded`

## Proposed fix

Per ONNX spec, Flatten with a 1-D input and `axis=0` produces `[1, d]`; with `axis=1` produces `[d]`. Neither is ambiguous. Relax the rank-check in `onnx-ir/src/node/flatten.rs` to accept rank 1 and emit either a `reshape` (static shape known from input + axis) or a direct pass-through in codegen.

## Refs

- #314 — parent scoreboard issue

kimjune01:
## Summary
- Fix Flatten node to accept rank-1 tensors as input, which is valid per the ONNX specification but previously caused a panic
- Move axis range validation from `extract_config` to `infer_types` to avoid panic path through `build_node`
- Add opset < 11 gate for negative axis values, matching the depth_to_space pattern
- Fix `reshape::<2>` codegen for axis=0 (was missing second type param `_`)
- Add rank-1 integration test with `.onnx` model (shape [5], axis=0)
- Advance 14 layer_normalization expanded tests from skip-codegen to skip-compile (they now pass the Flatten stage but fail later due to unrelated Shape/Size codegen issues)

Fixes #349

## Test plan
- 19 onnx-ir unit tests pass (including opset < 11 gate, axis range validation via infer_types)
- 5 burn-onnx codegen snapshot tests pass (updated for `reshape::<2, _>`)
- 3 onnx-tests integration tests pass: `flatten`, `flatten_2d`, and new `flatten_rank1`
- `cargo fmt --check` passes


crates/onnx-ir/src/node/flatten.rs
```diff
@@ -154,7 +152,15 @@ impl NodeProcessor for FlattenProcessor {
             axis += tensor.rank as i64;
         }
 
-        // TODO: Validate axis is within valid range [0, rank) after normalization - Invalid axis values should return error - Missing range validation
+        // Validate axis is within valid range [0, rank] after normalization.
+        // ONNX spec allows axis in [-r, r] inclusive; after normalization that's [0, rank].
+        if axis < 0 || axis as usize > tensor.rank {
+            return Err(ProcessError::Custom(format!(
```
antimora: This new error path is also reachable through `build_node` (graph.rs:202), which runs before `validate_node_spec` at type_inference.rs:119. `build_node` calls `extract_config().expect("Config extraction failed")`, so a malformed model with an out-of-range axis panics with no node name or axis value, instead of returning a clean `ProcessError`.

Cleanest fix is moving the axis range check into `infer_types` (which runs after validation), mirroring the rank-0 reject at line 72. The unit test would shift to calling `infer_types` instead of `extract_config`.

crates/onnx-ir/src/node/flatten.rs
```diff
@@ -154,7 +152,15 @@ impl NodeProcessor for FlattenProcessor {
...
+                "Flatten: axis {} out of range for input rank {} (valid range: [0, {}])",
+                axis, tensor.rank, tensor.rank
+            )));
+        }
+
         // TODO: Validate negative axis support for opset < 11 - Negative axis added in opset 11, should error for earlier opsets - Missing opset-specific validation
```
antimora: Real gap. Negative axis was added in opset 11 but the code accepts it for any opset >= 1. Other ops have the gate (depth_to_space.rs:100, max_pool*.rs). `_opset` is already in scope. Either implement the one-liner and drop both this TODO and the test TODO at line 394, or close them out with a note that we're intentionally permissive.

antimora (commented):
Thanks for picking this up.

A few high-level things before the inline notes:

1. Branch is forked at ad2fcb8a, but main has since merged #396 (version 0.21.0) and #400 (burn rev bump). The Cargo.toml/Cargo.lock diff here is stale and should drop on rebase.

2. PR body says the 14 layer_normalization expanded tests were promoted to `pass`, but commit 8480bf42 demoted them to `skip-compile`. They don't run, so they don't actually validate the rank-1 fix end to end. Worth updating the description.

3. No integration test for rank-1 in `crates/onnx-tests/tests/flatten/`. A small .onnx with shape [5] for axis=0 and axis=1 would lock the fix in.

Core logic looks right and all unit tests pass locally.

antimora (approved):
LGTM!

Thank you for your contribution!

PR #417, opened 2026-06-12
Add GlobalLpPool

### LINKED ISSUE #159: Implement GlobalLpPool operator
## Context

Gap analysis against [tract](https://github.com/sonos/tract).

## Operator

**GlobalLpPool** - computes Lp pool over all spatial dimensions.

- [ONNX spec](https://onnx.ai/onnx/operators/onnx__GlobalLpPool.html)
- Not currently in `node.rs`

## Attributes
- `p` (int, default 2) - the exponent of the Lp norm

anagaev:
### Checklist

It fails on an unrelated typo in `crates/onnx-ir/src/node/einsum.rs`. 

### Related Issues/PRs

- Closes #159 

### Changes
The PR add  `GlobalLpPool` operator.

- Add `burn-onnx` codegen
- Add `onnx-ir` node
- Register the operator
- Update `SUPPORTED-ONNX-OPS.md`

### Testing

- `cargo xtask validate`
- `cargo test -p onnx-ir global_lp_pool -- --nocapture`
- `cargo test -p burn-onnx global_lp_pool -- --nocapture`
- `cargo test -p onnx-tests global_lp_pool -- --nocapture`

### Additional note:
I’m not very familiar with the codebase and some concepts. So, there are a few things I’d like to highlight:

- I did not succeed in running `GlobalLpPool` operator with `opsetid=22` using either `onnxruntime` or `onnx.reference`.
- Because of the previous point, I didn't add support for Opset version 22 to `crates/onnx-ir/src/node/global_lp_pool.rs`
- In `crates/onnx-tests/tests/global_lp_pool/global_lp_pool.py` the values` ir_version=8` and `opset_version=7` are used. I’m not sure if this is correct.


crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+//! GlobalLpPool operation.
+//!
+//! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__GlobalLpPool.html>
+//!
+//! ## Type Constraints
+//! - T: tensor(double), tensor(float), tensor(float16)
+//!
+//! ## Opset Versions
+//! - **Opset 1**: Initial version (types: float16, float, double)
+//!   **Opset 2**: Initial version (types: float16, float, double)
```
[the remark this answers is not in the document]
anagaev: As I mentioned in the description I didn't succeed in running `GlobalLpPool` operator with `opsetid=22` using either `onnxruntime` or `onnx.reference`.  I'm not sure if adding opset 22 without a proper test would be a could idea

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+                tensor_ty.rank
+            )));
+        };
+
+        if !matches!(tensor_ty.dtype, DType::F16 | DType::F32 | DType::F64) {
+            return Err(ProcessError::TypeMismatch {
+                expected: "DType::F16 | DType::F32 | DType::F64".to_string(),
+                actual: format!("{:?}", tensor_ty.dtype),
+            });
+        }
```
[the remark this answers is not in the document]
anagaev: I think this issue is related to the previous one

crates/burn-onnx/src/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,176 @@
...
+        // scale/bias/mean/var are 1D tensors of shape [C], need to broadcast
+        // to match input shape [N, C, D1, D2, ...]
+        let dims: Vec<isize> = {
+            // For rank-D input, unsqueeze dim 0 and dims 2..rank-1
+            let mut dims = vec![]; // prepend batch dim
+            for i in 2..rank {
+                dims.push(i as isize);
+            }
+            dims
+        };
```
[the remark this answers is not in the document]
anagaev: I used implementation of other operators as references. So yes, it's copy-pasted
Does `onnx-ir` validate the input rank? In this case, this rank check could be removed, right?

anagaev:
@antimora

antimora:
> @antimora 

I am traveling. @laggui can help

anagaev:
@antimora @laggui could you please check the pr, if you have time?

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+            config,
+        })
+    }
+}
+
+fn extract_p(node: &RawNode) -> Result<i64, ProcessError> {
+    let p = node
+        .attrs
+        .get("p")
+        .map(|v| v.clone().into_i64())
```
laggui: The `NodeSpec` currently expresses support for `min_opset: 1` but it looks like in opset 1 `p` could be float, so this might panic. We can either change the min opset to 2, or update this.

crates/onnx-ir/src/node/global_lp_pool.rs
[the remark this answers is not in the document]
laggui: In this case, the `NodeSpec` below should probably set `max_opset: Some(21)`. And you can add a small unit test that exercises the unsupported opset error.

crates/onnx-tests/tests/global_lp_pool/global_lp_pool.py
```diff
@@ -0,0 +1,75 @@
...
+# ///
+
+
+import numpy as np
+import onnx
+import onnx.helper as helper
+from onnx import TensorProto
+import onnxruntime as ort
+from onnx.reference import ReferenceEvaluator
+import onnxruntime as ort
```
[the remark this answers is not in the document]
laggui: This should be fixed. Typically I'd say prefer using `onnx.reference.ReferenceEvaluator` instead of onnxruntime, as it is the standard guideline, but looks like other tests sometimes deviate from that. So either or, but just import one and use it 👍

laggui (commented):
Sorry for the delay! My notifications were flooded after I came back from leave 😅 I lost this one.

Just a couple of comments

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+pub(crate) struct GlobalLpPoolProcessor;
+
+impl NodeProcessor for GlobalLpPoolProcessor {
+    type Config = GlobalLpPoolConfig;
+
+    fn spec(&self) -> NodeSpec {
+        NodeSpec {
+            min_opset: 1,
```
antimora, line 44: GlobalLpPool isn't in the opset-compliance harness, so these opset claims aren't tested. Add it to `MIN_OPSET` in `tests/scripts/gen_rust_tests.py` and `SUPPORTED_OPS` in `tests/scripts/gen_opset_compliance.py`, then regenerate `tests/opset_compliance/`.

That gives parse tests for opsets 1, 2 and 22. The opset-1 one would have caught the `p` issue below.

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+                .clone()
+                .unwrap_or_else(|| vec![None; tensor_ty.rank]);
+
+            for el in shape.iter_mut().skip(2) {
+                *el = Some(1usize);
+            }
+            Some(shape)
+        };
+
+        extract_p(node)?;
```
antimora: This is load-bearing but untested.

Remove it and a `p=-1` model panics from `build_node`'s `.expect()` at line 110 instead of returning a clean error with the file and node name.

Remove it and all 14 tests still pass. The tests that reach `infer_types` all use `p=2` or no `p`, and the only negative-`p` test calls `extract_config` directly. So it reads as dead code and someone will eventually delete it.

Add an `infer_types` test for `p=-4` and `p=0`, and move the call up with the other validations. `lp_normalization.rs:91` has a comment explaining why it's there.

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+        if tensor_ty.rank <= 2 {
+            return Err(ProcessError::Custom(format!(
+                "GlobalLpPool: input tensor requires rank at least 3, got rank {}",
+                tensor_ty.rank
+            )));
+        };
+
+        if !matches!(tensor_ty.dtype, DType::F16 | DType::F32 | DType::F64) {
+            return Err(ProcessError::TypeMismatch {
+                expected: "DType::F16 | DType::F32 | DType::F64".to_string(),
```
antimora: This prints Rust syntax at the user:

```
Type mismatch: expected DType::F16 | DType::F32 | DType::F64, got I64
```

`lp_normalization.rs:86` uses `"Floating-point tensor dtype"`.

Also lines 58/61/67/130: `type_inference.rs:126` already prefixes `Node '{name}' ({node_type}):`, so the `"GlobalLpPool: "` prefixes double up. Line 61 puts a sentence in `expected` and comes out as `expected GlobalLpPool: input should be a tensor, got Shape(3)`. Elsewhere that field is just a type name.

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+
+fn extract_p(node: &RawNode) -> Result<i64, ProcessError> {
+    let p = node
+        .attrs
+        .get("p")
+        .map(|v| v.clone().into_i64())
+        .unwrap_or(2);
+
+    if p <= 0 {
+        return Err(ProcessError::Custom(format!(
```
antimora: Both this and the rank check at line 66 return `ProcessError::Custom`, so `test_global_lp_pool_extract_config_p_negative` can't tell which one fired. It asserts `Err(Custom(_))`, which the rank error satisfies too.

`InvalidAttribute { name, reason }` would let the test assert `name == "p"`, like `lp_normalization.rs:159`.

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,391 @@
...
+                expected: "DType::F16 | DType::F32 | DType::F64".to_string(),
+                actual: format!("{:?}", tensor_ty.dtype),
+            });
+        }
+
+        let static_shape = {
+            let mut shape = tensor_ty
+                .static_shape
+                .clone()
+                .unwrap_or_else(|| vec![None; tensor_ty.rank]);
```
antimora: `shape` takes its length from the input's `static_shape` rather than from `rank`. If those disagree, this keeps the wrong length and then writes `rank: tensor_ty.rank` next to it at line 95, producing a `TensorType` that contradicts itself.

`argument.rs:204` treats that mismatch as possible (returns `false` rather than asserting), so it isn't purely theoretical. Low priority, but seeding `vec![None; rank]` and copying only the first two entries is the same length and can't drift.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        // Build the reshape dimensions: [1, C, 1, 1, ...] for broadcasting
+        // scale/bias/mean/var are 1D tensors of shape [C], need to broadcast
+        // to match input shape [N, C, D1, D2, ...]
+        let dims: Vec<isize> = {
+            // For rank-D input, unsqueeze dim 0 and dims 2..rank-1
+            let mut dims = vec![]; // prepend batch dim
+            for i in 2..rank {
+                dims.push(i as isize);
+            }
```
antimora: To answer your question above: yes, onnx-ir validates rank. Line 65 rejects rank <= 2, which matches ORT (`Input dimension cannot be less than 3`). So this check is redundant today.

The failure mode is worth knowing though. `dims` is `2..rank`, so rank 2 gives an empty slice, and `sum_dims` folds over it, making it the identity. You'd get `input.abs()` for p=1 and `|x|` for p=2: compiles, runs, wrong numbers, no error.

So the panic at line 20 covers the case that would only crash, and nothing covers the case that silently corrupts. If you drop the panic, `debug_assert!(!dims.is_empty())` here is cheap.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,176 @@
...
+            dims
+        };
+        let mut body = quote! {let x = #input.abs();};
```
antimora: `abs()` does nothing when `p` is even. p=2 generates `x.abs().square()` and p=8 generates `x.abs().powi_scalar(8)`, so it's a wasted pass over the input, including on the default path.

burn splits on the same thing:

```rust
/// Computes ``L(2*n)`` for even integer ``n``.
///
/// This lets us skip the abs.
fn lp_signed_norm<B: Backend, const D: usize>(x: Tensor<B, D>, p: u32, dim: usize) -> Tensor<B, D> {
    x.powi_scalar(p).sum_dim(dim).powf_scalar(1. / (p as f64))
}
```

Only emit it for odd `p`.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,176 @@
...
+        match p {
+            1 => {
+                body.extend(quote! {x.sum_dims(&[#(#dims),*])});
+            }
+            2 => {
+                body.extend(quote! {x.square().sum_dims(&[#(#dims),*]).sqrt()});
+            }
+            _ => {
+                body.extend(
+                    quote! {x.powi_scalar(#p).sum_dims(&[#(#dims),*]).powf_scalar(#inv_p) },
```
antimora: `GlobalLpPoolConfig::new` and `GlobalLpPoolNodeBuilder` are public and don't validate, so `extract_p`'s `p > 0` check only covers the pipeline path. The snapshot tests in this file already build configs directly. Through `codegen_forward_default`:

- `p = 0` makes `inv_p` infinite and panics inside proc-macro2 with `assertion failed: f.is_finite()`, which names neither the op nor the attribute.
- `p = -2` doesn't fail at all. It emits `x.powi_scalar(-2i64).sum_dims(&[2isize]).powf_scalar(-0.5f64)`, which compiles and is wrong.

`forward` can't return an error, so `unreachable!` is the option here, like `lp_normalization.rs:23`.

Separately: `powi_scalar` converts the exponent to the tensor's dtype, so `|x|^p` runs in f16 for f16 input. f16 caps at 65504, so p=8 overflows for `|x| > 4`.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,176 @@
...
+            .input_tensor("input", 3, DType::F32)
+            .output_tensor("output", 3, DType::F32)
+            .config(GlobalLpPoolConfig::new(4))
+            .build();
+        let code = codegen_forward_default(&node);
+        assert_snapshot!(code, @r"
+        pub fn forward(&self, input: Tensor<3>) -> Tensor<3> {
+            let output = {
+                let x = input.abs();
+                x.powi_scalar(4i64).sum_dims(&[2isize]).powf_scalar(0.25f64)
```
antimora: Also worth noting: the general arm is only snapshotted at p=4 and p=8. Both are powers of two, so `inv_p` is exact (`0.25f64`, `0.125f64`).

p=3 is what the integration models use, and it's the one that gives `0.3333333333333333f64`. Changing this to `new(3)` fixes the name and covers that.

crates/onnx-tests/tests/global_lp_pool/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    global_lp_pool_rank_4_l3
+);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use burn::tensor::{Device, Tensor, TensorData, Tolerance};
+
+    // Input generated via np.random.seed(42), shape [2, 3, 4]. Shared across all
+    // tests so different axis/p configurations operate on the same data.
```
antimora: Copied from `lp_normalization/mod.rs:14-15`, `axis/p` included. GlobalLpPool has no `axis` attribute.

"Shared across all tests" isn't right either, now that `global_lp_pool_input_rank_4_l3` uses `test_input_rank_4`.

`test_input_rank_4` at line 35 has no comment at all, worth adding the seed and shape.

crates/onnx-tests/tests/global_lp_pool/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+            [[4.4424477], [3.6474063], [3.1837015]],
+        ]);
+
+        output
+            .to_data()
+            .assert_approx_eq::<f32>(&expected, Tolerance::default());
+    }
+
+    #[test]
+    fn global_lp_pool_l2() {
```
antimora, line 113: All five models are opset 7, float32, `p` as INT.

This and `global_lp_pool_default` assert identical tensors. That's correct since `p` defaults to 2, but it means one of them only tests that the attribute is optional.

Rank 4 is only tested at p=3, so p=1 and p=2 never run end to end at rank 4. That's the gap I'd fill first. Nothing covers f16 or f64 either, though `infer_types` accepts both and `test_utils.rs` has the helpers.

antimora (commented):
Reviewed at 35d00a5e. Skipping what's already been raised (opset-1 float `p`, the bfloat16 gate and docs, the batch-norm comment block, the `l3`/p=4 name, the Python imports).

Checked and correct:

- Expected values are real. I recomputed all five from `sum(|x|^p)^(1/p)` in NumPy, and the committed `.onnx` files match the generator script.
- `rank >= 3` matches ORT.
- `static_shape` inference is right, and more thorough than GlobalAveragePool next to it.
- Registration is complete in all seven places.
- Opset 22 already works. I converted an opset-22 f32 model with onnx2burn successfully.

On `max_opset: Some(21)`: GlobalLpPool-22 is the same as v2 apart from bfloat16 in `T`, so capping at 21 would reject opset-22 models that work today. Gating just the dtype on `opset >= 22`, like `lp_normalization.rs:76-83`, avoids that.

Unrelated to this PR: `lp_pool1d.rs:52` and `lp_pool2d.rs:50` have the same issue, `min_opset: 1` with `into_i64()` on `p`, and LpPool's `p` is also FLOAT at opset 1. Probably worth its own issue.

crates/onnx-ir/src/node/global_lp_pool.rs
anagaev: I'm not sure I fully understand a few points here.

The input's `static_shape` and `rank` can disagree, right? Given the previous point and how `argument.rs` `merge_static_shape` works, is it better to use `rank` rather than `static_shape`? So, would defining `static_shape` as `vec![None; tensor_ty.rank]` and copying the first two values from the input's `static_shape`, if available, be the best approach here?

Out of curiosity, what are the possible practical situations where the input's `static_shape` and `rank` would disagree? I thought they would always agree, and that `static_shape` was used only to store additional information when it was available.

crates/onnx-ir/src/node/global_lp_pool.rs
anagaev: It seems this part isn't covered by the development guide.
So I added locally `GlobalLpPool: 1` to `MIN_OPSET` in `tests/scripts/gen_rust_tests.py` and  updated `tests/scripts/gen_opset_compliance.py` (`SUPPORTED_OPS`, `GENERATORS` etc.). 
`gen_opset_compliance.py` regenerated correctly some `*.onnx` in `tests/fixtures/opset_compliance`. But `gen_rust_tests.py` generated `tests/opset_compliance` without reference values for 
`insta::assert_snapshot!` in tests. Is it supposed to be like this?

antimora:
@anagaev, i'll take over and complete the feedback fixes; full credit to you; just making sure we complete this before the upcoming release.

crates/onnx-ir/src/node/global_lp_pool.rs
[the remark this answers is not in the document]
antimora: Fixed the docs to list bfloat16 and to say what each opset actually changed:

```
//! ## Opset Versions
//! - **Opset 1**: Initial version (types: float16, float, double). `p` is a FLOAT attribute.
//! - **Opset 2**: `p` becomes an INT attribute.
//! - **Opset 22**: Adds bfloat16 to T.
```

On `max_opset: Some(21)`: I left it as `None`. GlobalLpPool-22 is identical to v2 apart from bfloat16 in `T`, so capping at 21 would reject opset-22 f32 models that already convert correctly today. The dtype alone is gated on the opset instead, matching `lp_normalization.rs:76-83`, and `test_global_lp_pool_bf16_gated_on_opset` asserts both sides (bf16 rejected at 21, accepted at 22).

@anagaev on not being able to run opset 22 end to end: parsing is now covered by the opset-compliance harness, which generates an opset-22 model for this op.

crates/onnx-ir/src/node/global_lp_pool.rs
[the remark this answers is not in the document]
antimora: Fixed. Type inference gates the dtype allowlist on the opset rather than rejecting BF16 outright:

```rust
// bfloat16 joins T at opset 22.
let allowed = if opset >= 22 {
    matches!(tensor_ty.dtype, DType::BF16 | DType::F16 | DType::F32 | DType::F64)
} else {
    matches!(tensor_ty.dtype, DType::F16 | DType::F32 | DType::F64)
};
```

`test_global_lp_pool_bf16_gated_on_opset` covers rejection at opset 21 and acceptance at 22.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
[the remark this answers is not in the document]
antimora: Fixed both.

The comment block was copied from a batch-norm node and described a `[1, C, 1, 1]` broadcast reshape this op never performs. It now describes what the code computes, and the loop it annotated collapsed into a `collect`.

The panic is now a `compile_error!` carrying the node name, the shape ScatterElements uses:

```rust
_ => {
    let msg = format!("GlobalLpPool node '{}': input must be a tensor", self.name);
    return quote! { let #output = { compile_error!(#msg); unreachable!() }; };
}
```

@anagaev to your question: yes, onnx-ir validates rank, it rejects rank <= 2 which matches ORT. I kept the match here because it is how `rank` is obtained, but see the rank-check thread for the `debug_assert` that covers the case which would otherwise be silently wrong.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,176 @@
...
+        ");
+    }
+
+    #[test]
+    fn global_lp_pool_rank3_l3() {
+        let node = GlobalLpPoolNodeBuilder::new("global_lp_pool")
+            .input_tensor("input", 3, DType::F32)
+            .output_tensor("output", 3, DType::F32)
+            .config(GlobalLpPoolConfig::new(4))
+            .build();
```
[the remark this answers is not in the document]
antimora: Fixed by changing the config rather than the name, per the later thread: the test is `global_lp_pool_rank3_l3` with `p = 3`.

p=3 is what the integration models use, and it is the only case where `inv_p` is not exactly representable, so the snapshot now pins `0.3333333333333333f64`. p=4 and p=8 were both powers of two.

crates/onnx-tests/tests/global_lp_pool/global_lp_pool.py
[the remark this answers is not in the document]
antimora: Fixed the duplicate `onnxruntime` import and the unused one.

On preferring `ReferenceEvaluator`: it cannot run this operator. `onnx==1.19.0` has no GlobalLpPool implementation and raises

```
onnx.reference.op_run.RuntimeImplementationError: No registered implementation for
operator 'GlobalLpPool' and domain ''
```

GlobalAveragePool and GlobalMaxPool are implemented, GlobalLpPool is not. So the script keeps onnxruntime as the single evaluator with a comment recording why.

@laggui that is the "either or" you allowed, just noting the reason it cannot be the standard one here. As a cross-check I recomputed all seven expected tensors in NumPy from `sum(|x|^p)^(1/p)`, independently of both runtimes, and they match.

crates/onnx-ir/src/node/global_lp_pool.rs
antimora: Good catch, and it was a real panic rather than just a risk. `p` is FLOAT in opset 1 and INT from opset 2 on, and `into_i64()` panics rather than returning an error on a type mismatch.

I kept `min_opset: 1` and taught `extract_p` to accept both encodings:

```rust
let p = match node.attrs.get("p") {
    None => 2,
    Some(AttributeValue::Int64(p)) => *p,
    Some(AttributeValue::Float32(p)) => {
        if !p.is_finite() || p.fract() != 0.0 {
            return Err(ProcessError::InvalidAttribute {
                name: "p".to_string(),
                reason: format!("expected a whole number, got {p}"),
            });
        }
        *p as i64
    }
    Some(other) => { /* InvalidAttribute */ }
};
```

Matching on the variant rather than on the opset also covers exporters that write the other encoding for their declared opset.

Verified the new opset-1 compliance test is load-bearing: reverting `extract_p` to `into_i64()` fails it with

```
panicked at crates/onnx-ir/src/ir/attribute.rs:365:
Expected Int64, got Float32(2.0)
```

Fractional `p` is rejected rather than truncated, since codegen emits an integer exponent (`test_global_lp_pool_opset1_fractional_p_rejected`).

crates/onnx-ir/src/node/global_lp_pool.rs
antimora: Done. Added `"GlobalLpPool": 1` to `MIN_OPSET` in `gen_rust_tests.py`, `"GlobalLpPool": "global_lp_pool"` to `SUPPORTED_OPS` plus a `make_global_lp_pool` generator in `gen_opset_compliance.py`, and regenerated.

The generator emits `p` in the encoding each opset declares, which is what makes the opset-1 case load-bearing:

```python
# `p` is a FLOAT attribute in opset 1 and an INT from opset 2 on.
p_attr = 2.0 if opset < 2 else 2
```

Total goes 476 -> 479 (opsets 1, 2 and 22), still 0 failing.

And you were right that it would have caught the `p` issue. Reverting the parser to `into_i64()` fails `opset_compliance::opset_01::global_lp_pool` with `Expected Int64, got Float32(2.0)`.

crates/onnx-ir/src/node/global_lp_pool.rs
antimora: Fixed. `extract_p` now runs right after the dtype check, next to the other validations, with the same comment `lp_normalization.rs` uses:

```rust
// Validate here so malformed graphs fail before codegen.
extract_p(node)?;
```

And it is now reachable from a test through `infer_types`, not only through `extract_config`:

```rust
#[test]
fn test_global_lp_pool_infer_types_rejects_non_positive_p() {
    for bad in [-4, 0] { ... }
}
```

so deleting the call fails a test instead of silently moving the failure to `build_node`'s `.expect()`.

crates/onnx-ir/src/node/global_lp_pool.rs
antimora: Fixed. The dtype error now reads `Floating-point tensor dtype`, matching `lp_normalization.rs:86`, and all four `"GlobalLpPool: "` prefixes are gone since `phases/type_inference.rs:129` already wraps errors with `Node '{name}' ({node_type}):`.

The non-tensor error also puts a type name in `expected` rather than a sentence, so it now reads

```
Node 'x' (GlobalLpPool): Type mismatch: expected Tensor, got Shape(3)
```

instead of the doubled-up version.

crates/onnx-ir/src/node/global_lp_pool.rs
antimora: Fixed. `extract_p` returns `InvalidAttribute { name: "p", reason }` like `lp_normalization.rs:159`, and the test asserts the name so the rank error can no longer satisfy it:

```rust
assert!(matches!(
    processor.extract_config(&node, 16),
    Err(ProcessError::InvalidAttribute { ref name, .. }) if name == "p"
));
```

crates/onnx-ir/src/node/global_lp_pool.rs
antimora: Fixed by seeding from `rank`, which is what you proposed:

```rust
// Length comes from `rank`, so it cannot drift from the rank written below
// if the input's `static_shape` disagrees with its own rank.
let mut static_shape = vec![None; tensor_ty.rank];
if let Some(input_shape) = &tensor_ty.static_shape {
    for (out, inp) in static_shape.iter_mut().zip(input_shape).take(2) {
        *out = *inp;
    }
}
for el in static_shape.iter_mut().skip(2) {
    *el = Some(1usize);
}
```

@anagaev on when they disagree in practice: mostly they do not, and you are right that `static_shape` is supplemental information. The reason not to take the length from it is that nothing enforces the invariant. `TensorType` lets the two fields be set independently, `argument.rs:204` treats a length mismatch as a normal `false` return rather than asserting it away, and value_info in a model file is author-supplied and not guaranteed to agree with the declared rank. So it is less "this happens often" and more "when it does, the old code produced a `TensorType` that contradicted itself, and this version cannot".

Added `test_global_lp_pool_static_shape_shorter_than_rank` to pin it: rank 4 with a 2-element `static_shape` now yields `[2, 3, 1, 1]` rather than a length-2 shape sitting next to `rank: 4`.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
antimora: Kept the match, since it is how `rank` is obtained, but the panic is now a `compile_error!` naming the node.

Added the `debug_assert` you suggested for the case that would otherwise be silently wrong:

```rust
// onnx-ir rejects rank <= 2; an empty slice here would make `sum_dims` fold over
// nothing and act as the identity, which compiles and runs but returns the wrong
// numbers.
let dims: Vec<isize> = (2..rank).map(|i| i as isize).collect();
debug_assert!(!dims.is_empty(), "GlobalLpPool requires rank >= 3");
```

crates/burn-onnx/src/burn/node/global_lp_pool.rs
antimora: Fixed. `abs` is only emitted for odd `p`:

```rust
// |x| is redundant for even p, which raises away the sign anyway.
let x = if p % 2 == 0 { quote! { #input } } else { quote! { #input.abs() } };
```

The snapshots show it. p=2 is now

```rust
let x = input;
x.square().sum_dims(&[2isize]).sqrt()
```

and p=8 drops it too. p=1 and p=3 keep it.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
antimora: Fixed with `unreachable!`, matching `lp_normalization.rs:23`:

```rust
// onnx-ir rejects p <= 0, so this is only reachable from a hand-built node:
// `GlobalLpPoolConfig::new` and the node builder are public and don't validate.
// p = 0 would make `inv_p` infinite and panic inside proc-macro2, and a
// negative p emits code that compiles and is silently wrong.
if p <= 0 {
    unreachable!("p must be > 0, got {p}");
}
let inv_p = 1.0f64 / p as f64;
```

so p=0 names the op and the attribute instead of surfacing as `assertion failed: f.is_finite()` from inside proc-macro2, and p=-2 fails instead of emitting code that compiles and is wrong.

On the f16 point, I left the numerics alone and documented it instead. `powi_scalar` evaluating in the input dtype is also what ORT does, so upcasting to f32 for the power would make us diverge from the reference on exactly the models it would be meant to help. There is a comment at the call site recording the overflow bound. Happy to open a separate issue if you would rather it be handled.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
antimora: Done, as `new(3)` rather than a rename. The snapshot now pins the inexact reciprocal:

```rust
x.powi_scalar(3i64).sum_dims(&[2isize]).powf_scalar(0.3333333333333333f64)
```

p=8 stays as the even, power-of-two case, which now also covers the skipped `abs`.

crates/onnx-tests/tests/global_lp_pool/mod.rs
antimora: Fixed. Dropped the `axis` reference (copied from `lp_normalization`, GlobalLpPool has no such attribute) and the stale "shared across all tests", and gave `test_input_rank_4` the seed and shape it was missing:

```rust
// Input generated via np.random.seed(42), shape [2, 3, 4]. Shared by the rank-3
// tests so every p operates on the same data.

// Input generated via np.random.seed(42), shape [2, 3, 2, 3]. Shared by the
// rank-4 tests.
```

crates/onnx-tests/tests/global_lp_pool/mod.rs
antimora: Filled the rank-4 gap first, as you suggested. Added `global_lp_pool_rank_4_l1` and `global_lp_pool_rank_4_l2` models and tests, so rank 4 now runs end to end at p=1, 2 and 3.

Also recorded why `default` and `l2` assert the same values:

```rust
// `p` defaults to 2, so this asserts the same values as `global_lp_pool_l2`.
// What it actually covers is that the attribute is optional.
```

f16 and f64 are covered at the `infer_types` level (`test_global_lp_pool_accepts_f16_and_f64`) rather than end to end. The generator writes `TensorProto.FLOAT` graphs and the expected values are f32 literals, so an f16 end-to-end model needs its own tolerance story. Say the word if you want that in this PR rather than a follow-up.

All expected values were recomputed in NumPy independently of onnxruntime.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
antimora: Follow-up: I've replaced this `debug_assert!` with a `compile_error!`.

What changed my mind is the argument you made in this thread: the empty-`dims` case is the one that silently corrupts rather than crashing. `debug_assert!` compiles out in release, so it was the weakest guard sitting on the worst failure mode; a rank-2 hand-built node would have generated an identity reduction in a release build with no diagnostic at all.

It now emits

```
compile_error!("GlobalLpPool node 'global_lp_pool': requires rank >= 3, got rank 2");
```

with a snapshot test covering it.

crates/burn-onnx/src/burn/node/global_lp_pool.rs
antimora: Follow-up: I've switched this from `unreachable!` to `compile_error!`.

Copilot's re-review flagged the branch again after the push, on the grounds that it is reachable from the public builder so it should not crash model generation. Two of my own cleanup passes had independently flagged the same inconsistency, since the non-tensor branch above it already emitted `compile_error!`.

It now emits

```
compile_error!("GlobalLpPool node 'global_lp_pool': p must be finite and > 0, got 0");
```

instead of panicking, with a snapshot test. That departs from the `lp_normalization.rs:23` precedent you pointed me at, so say the word if you would rather keep the panic and I will put it back.

crates/burn-onnx/src/import/burn/node/global_lp_pool.rs
```diff
@@ -0,0 +1,206 @@
...
+            }
+        };
+
+        // onnx-ir rejects p <= 0, so this is only reachable from a hand-built node:
+        // `GlobalLpPoolConfig::new` and the node builder are public and don't validate.
+        // p = 0 would make `inv_p` infinite and panic inside proc-macro2, and a
+        // negative p emits code that compiles and is silently wrong.
+        if p <= 0 {
+            unreachable!("p must be > 0, got {p}");
+        }
```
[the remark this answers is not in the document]
antimora: Agreed, and both are now `compile_error!`.

Two of my own cleanup passes had flagged the same thing independently: three different escalation mechanisms guarding one class of condition inside a 50-line function, when the non-tensor branch right above them already emitted `compile_error!`.

The release-build half of your point is the sharper one. `debug_assert!` compiles out, so a rank-2 hand-built node silently generated an identity reduction in release: `dims` is empty, and `sum_dims` over an empty slice folds over nothing. That is the worst of the three failure modes and it had the weakest guard.

All three preconditions now go through one helper:

```rust
let invalid = |reason: String| {
    let msg = format!("GlobalLpPool node '{}': {reason}", self.name);
    quote! { let #output = { compile_error!(#msg); unreachable!() }; }
};
```

with a snapshot test each:

```
compile_error!("GlobalLpPool node 'global_lp_pool': requires rank >= 3, got rank 2");
compile_error!("GlobalLpPool node 'global_lp_pool': p must be finite and > 0, got 0");
```

@antimora this reverses the `unreachable!` / `debug_assert!` shape you asked for on the two threads above. Flagging it rather than changing it quietly; I've left a note on both.

crates/onnx-ir/src/node/global_lp_pool.rs
```diff
@@ -0,0 +1,539 @@
...
+        None => 2,
+        Some(AttributeValue::Int64(p)) => *p,
+        // Opset 1 stores `p` as a float. Only whole values map onto the integer
+        // exponent codegen emits, so reject the rest rather than truncating.
+        Some(AttributeValue::Float32(p)) => {
+            if !p.is_finite() || p.fract() != 0.0 {
+                return Err(ProcessError::InvalidAttribute {
+                    name: "p".to_string(),
+                    reason: format!("expected a whole number, got {p}"),
+                });
```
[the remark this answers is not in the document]
antimora: You're right, and it's now supported rather than rejected.

I checked the schema registry before changing anything:

```
GlobalLpPool-1   p -> FLOAT, default 2.0   (no integrality constraint)
GlobalLpPool-2   p -> INT,   default 2
```

so `p = 2.5` at opset 1 is spec-legal, and `sum(|x|^p)^(1/p)` is well defined for it.

`GlobalLpPoolConfig::p` is now `f64`, which carries both encodings losslessly, and codegen picks the exponent form:

```rust
let whole = (p.fract() == 0.0).then(|| p as i64);
let reduced = match whole {
    Some(1) => quote! { x.sum_dims(&#dims) },
    Some(2) => quote! { x.square().sum_dims(&#dims).sqrt() },
    Some(n) => quote! { x.powi_scalar(#n).sum_dims(&#dims).powf_scalar(#inv_p) },
    None => quote! { x.powf_scalar(#p).sum_dims(&#dims).powf_scalar(#inv_p) },
};
```

Integer exponents keep `powi_scalar`, which has multiply fast paths for small p. The `abs` skip is now conditional on `whole` being even, because a fractional exponent always needs it: `powf_scalar` of a negative base is NaN.

```rust
pub fn forward(&self, input: Tensor<3>) -> Tensor<3> {
    let output = {
        let x = input.abs();
        x.powf_scalar(2.5f64).sum_dims(&[2]).powf_scalar(0.4f64)
    };
    output
}
```

`test_global_lp_pool_opset1_fractional_p_rejected` became `test_global_lp_pool_opset1_fractional_p`, as you suggested, and there is an end-to-end model as well (`global_lp_pool_opset1_fractional_p.onnx`, opset 1, p = 2.5).

Worth recording for that one test: neither reference runtime can execute it. onnxruntime has no GlobalLpPool(1) kernel at all,

```
NOT_IMPLEMENTED : Could not find an implementation for GlobalLpPool(1) node
```

which is independent of `p` (it rejects p = 2.0 the same way), and `onnx.reference` has no GlobalLpPool implementation for any opset. So that model's expected values come from the spec formula computed in NumPy, which the generator does explicitly and comments. Every other model still goes through onnxruntime.

Non-finite `p` is still rejected, since neither NaN nor infinity yields a usable exponent (`test_global_lp_pool_rejects_non_finite_p`).

anagaev:
@antimora Thank you for your fixes and explanations! I’ll keep them in mind for my future contributions

PR #439, opened 2026-08-04
Update burn to 0.22.0-pre.1

laggui:
Preparing burn-onnx pre-release


laggui:
All checks passed; though there is a known regression as highlighted in https://github.com/tracel-ai/burn/issues/5219

We're still moving forward with the pre-release here but this should be addressed before the next major release.

PR #454, opened 2026-08-17
feat(export): add ONNX export from captured Burn modules

### LINKED ISSUE #22: Export Burn model to ONNX
It would be nice too export a burn model to onnx format for use in other DL frameworks, and, in my use case, use an onnx model for inference when inference in specific domains is required and there already exist optimization for onnx models.

laggui:
**Disclaimer:** not all operations are currently supported for export, this will be incremental. The first target was a vision model.

### Checklist

- [ ] Confirm relevant documentation has been updated.
- [ ] For new/modified ONNX operators: verified implementation matches `onnx-spec/ops/<OpName>.md`.

### Related Issues/PRs

Requires https://github.com/tracel-ai/burn/pull/5377
Closes #22 

### Changes

Added `OnnxExporter` that captures a model's forward pass (`GraphIr`) and returns the ONNX model.

Example usage:
```rust
let module = ResNet18::new(10, &device);
let model = OnnxExporter::new()
    .export(&module, input, ResNet18::forward)
    .unwrap();
```

*Note: it doesn't share existing importer code because the pipelines have opposite responsibilities: import transforms ONNX into a codegen-oriented representation, while export lowers captured burn `GraphIr` directly into ONNX protobuf.*

### Testing

Unit + integration tests


antimora:
In the future it'd be nice if GraphIr is translated into onnx-ir graph and onnx-ir would be responsible for creating and dealing with protobuf/opset stuff.

laggui:
I initially looked into reusing onnx-ir, but the current `OnnxGraph` is a processed, importer-oriented representation rather than a neutral writable ONNX IR. So for the initial exporter I preferred going with a focused implementation that mapped the `GraphIr` (burn-ir) -> `ResolvedExportGraph` (burn-onnx) -> ONNX protobuf (onnx-ir).

In the future we could try to extract a smaller bidirectionnal ONNX model layer that both the import and export can use. But I believe its current form will have to be adapted for that.

crates/burn-onnx/src/export/lower/module.rs
```diff
@@ -0,0 +1,206 @@
...
+                vec![output],
+            );
+            context.float_attribute("epsilon", batch_norm.epsilon.elem::<f32>());
+            Ok(true)
+        }
+        ModuleOperationIr::Interpolate(interpolate) => {
+            let (mode, coordinate_mode, nearest_mode) = match interpolate.options.mode {
+                InterpolateModeIr::Nearest => ("nearest", "asymmetric", Some("floor")),
+                InterpolateModeIr::NearestExact => {
+                    ("nearest", "half_pixel", Some("round_prefer_floor"))
```
antimora: `round_prefer_floor` is round-half-down, so this is off by one whenever `(d + 0.5) * in/out` lands exactly on an integer.

Burn's nearest-exact is `floor((d + 0.5) * in/out)`. With `half_pixel` ONNX computes `x_orig = y - 0.5`, which is an exact half-integer precisely when `y` is an integer, and round-half-down then picks one index too low. For an integer downsample ratio `y` is *always* integral, so every pixel is wrong:

```
4 -> 2 nearest-exact:  burn reads rows [1, 3],  exported model reads [0, 2]
```

`round_prefer_ceil` (round half up) gives `floor(y)` in both the integral and non-integral cases, so it matches burn everywhere.

Upsampling is unaffected because `y` is never integral there, which is why `interpolate_matches_burn` misses this. Worth adding a 2x downsample case with the fix.

crates/burn-onnx/src/export/lower/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+    Ok(())
+}
+
+fn value_info(
+    tensor: &TensorIr,
+    dynamic_axes: &[crate::export::DynamicAxis],
+) -> Result<ValueInfoProto, ExportError> {
+    let mut info = ValueInfoProto::new();
+    info.name = name(tensor.id);
```
antimora: This is the only naming site that doesn't go through `tensor_name`, so a graph output that is also a captured parameter gets declared as `tensor_N` while its data lives in an initializer named `weight`.

```rust
fn forward(&self, x: Tensor<1>) -> (Tensor<1>, Tensor<1>) {
    (x + self.weight.val(), self.weight.val())
}
```

exports Ok with `outputs = ["tensor_7", "tensor_2"]` and `initializers = ["weight"]`. Nothing produces `tensor_2`, so ORT rejects it at session creation.

`validate_bindings` already forbids an initializer name on a runtime *input*; the output side is unguarded.

crates/burn-onnx/src/export/exporter.rs
```diff
@@ -0,0 +1,572 @@
...
+impl<T: sealed::SealedExportInput> ExportInput for T {}
+
+macro_rules! impl_tensor_value {
+    ($kind:ty) => {
+        impl<const D: usize> sealed::SealedExportValues for Tensor<D, $kind> {
+            fn collect_tensor_ids(&self, ids: &mut Vec<TensorId>) {
+                let primitive = self
+                    .clone()
+                    .try_into_primitive::<CaptureBackend>()
+                    .expect("export tensor must be on the capture device");
```
antimora: This one is reachable from ordinary user code. `collect_tensor_ids` runs on the result of the forward closure, and that result never goes through `to_capture_device`, so

```rust
|_, x| (x.clone(), Tensor::zeros([2], &Device::default()))
```

panics with `BackendMismatch("Expected Capture tensor, got variant: Flex")` instead of returning an error. Building a tensor on the default device inside a forward pass is a normal thing to write.

`collect_tensor_ids` is on a sealed private trait, so returning `Result<(), ExportError>` here costs nothing in public API. Same for the three `ParameterNameVisitor` sites below.

crates/burn-onnx/src/export/lower/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+/// Lower a resolved graph with concrete values and stable initializer names.
+///
+/// `initializer_names` typically maps captured module parameter tensor IDs to
+/// their module paths. Unnamed initialized values retain deterministic tensor-ID names.
+pub fn export_graph_with_bindings(
+    graph: &ResolvedExportGraph,
+    values: &BTreeMap<TensorId, TensorData>,
+    runtime_inputs: &[TensorId],
+    initializer_names: &HashMap<TensorId, String>,
```
antimora: `HashMap` here is `hashbrown::HashMap` (line 24), and hashbrown is a private dep that isn't re-exported. A downstream caller can't build this argument without adding hashbrown to their own manifest and matching your minor version.

Either take `std::collections::HashMap` at the boundary or make this `pub(crate)`. Nothing outside the crate calls it today.

crates/burn-onnx/src/export/shape/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+                    axis,
+                    reason: format!("operation {index} has varying slice bounds"),
+                });
+            }
+            (
+                OperationIr::BaseFloat(BaseOperationIr::SliceAssign(sample_slice))
+                | OperationIr::BaseInt(BaseOperationIr::SliceAssign(sample_slice)),
+                OperationIr::BaseFloat(BaseOperationIr::SliceAssign(validation_slice))
+                | OperationIr::BaseInt(BaseOperationIr::SliceAssign(validation_slice)),
+            ) if sample_slice.ranges != validation_slice.ranges => {
```
antimora, line 509: The pad-equality check below is gated on the ranges differing, so it never runs in the case where the ranges match but the `Full` output shape doesn't.

Padding a statically-shaped tensor on the right by an amount derived from some *other* dynamic input gives identical ranges in both traces with different `Full` shapes. `constant_pad` computes `pads` from `slice_assign.tensor.shape`, so the sample's pad amount is baked in silently.

The doc comment above this function says lowering can never embed a sample-only value, which is exactly the right invariant to state. This is the gap in it.

crates/burn-onnx/src/export/shape.rs
```diff
@@ -0,0 +1,726 @@
...
+                }
+                symbols.sort();
+                symbols.dedup();
+                dynamic_axes.push(DynamicAxis {
+                    tensor: sample_id,
+                    axis,
+                    symbol: if symbols.len() == 1 {
+                        symbols.pop().unwrap()
+                    } else {
+                        format!("output_{position}_dim_{axis}")
```
antimora: When two dynamic axes happen to share the same `(sample, validation)` pair, this quietly invents a fresh symbol and drops the relationship.

Square inputs hit it right away: `[1,1,2,2]` / `[2,1,4,4]` with `height` and `width` both dynamic gives both axes `(2, 4)`, so the output is declared `output_0_dim_2`, `output_0_dim_3` instead of `height`, `width`. Exports fine, and anything downstream doing shape inference loses the input/output relation.

The reshape-operand path below treats the identical ambiguity as a hard `DynamicShapeLost`. Would be nice for the two to agree. Rejecting two distinct symbols that share a dim pair in `validate_input_specs` would also make the failure obvious at the call site.

crates/burn-onnx/src/export/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+mod exporter;
+mod lower;
+mod model;
+mod resolved;
+mod shape;
+mod validate;
+
+pub use error::ExportError;
+pub use exporter::{ExportInput, ExportValues, OnnxExporter, Opset};
+pub use lower::{
```
antimora: Worth trimming this list before it ships. Of the 17 items exported here, only `OnnxExporter`, `Opset`, `InputSpec`, `AxisSpec`, `OnnxModel` and `ExportError` have any caller outside `src/export/`.

The other 11 are also the ones that leak burn internals: `ResolvedExportGraph.graph` exposes `GraphIr`, and `ShapeExpr` / `DynamicAxis` / `ExportError` expose `TensorId`. There's no `#[non_exhaustive]` anywhere in `burn-ir`, so an `OperationIr` variant added upstream turns into a semver break here.

Demoting them to `pub(crate)` is free right now and breaking once 0.22 is out. Easy to promote later if someone actually needs the low-level entry points.

antimora (commented):
Went through the export path in detail. The overall shape of this is good: `lower_operation` is built so an unsupported op can't silently fall through, `Lanczos3` and non-`[1,1]` `AdaptiveAvgPool2d` refuse rather than approximate, and pairing `erase_shape_sensitive_attributes` with `validate_shape_sensitive_operations` (and saying so in a comment) is the right instinct for a tracing exporter. Conv/pool `pads` ordering, bilinear `half_pixel`, and the IR-8/opset-18 pairing all check out.

Built and ran it locally: `--features export` and `--features import` both compile clean, and the 33 tests pass. The ORT numerical tests comparing against the Burn module are a much better oracle than structural assertions.

One real bug (the `NearestExact` rounding mode), plus a few things worth settling before the public API is locked in. Comments inline.

Two non-inline notes:

- `Cargo.toml` pins burn to `e3b258cee`, which is only on `upstream/fix/capture/initializers`. Presumably waiting on tracel-ai/burn#5377 to land.
- Nothing in CI builds `import` and `export` together, though `[package.metadata.docs.rs]` lists `["default", "export"]`. I checked and it does compile today, so this is just a coverage gap. Clippy also never sees `src/export/` since `xtask check lint` uses default features.

crates/burn-onnx/src/export/exporter.rs
```diff
@@ -0,0 +1,572 @@
...
+            .map_err(|error| ExportError::InvalidBoundary(error.to_string()))?;
+        let Some((input_ids, parameter_names)) = capture_metadata else {
+            return Err(ExportError::InvalidBoundary(
+                "capture scope did not execute".into(),
+            ));
+        };
+        captured
+            .graph
+            .operations
+            .retain(|operation| !matches!(operation, OperationIr::Init(_)));
```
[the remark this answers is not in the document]
antimora: Confirmed. After the `retain`, `|_, x| x` leaves `operations` empty, and both `find_tensor` and `shape.rs::tensor` only scan operation nodes, so the boundary lookup at the top of `export_graph_with_bindings` fails with `MissingValue`.

Worth noting the error is unhelpful on its own here: you get `missing value for tensor 6` with no hint that the cause is a pass-through input. `MissingValue` carrying the input position would make this a lot easier to act on.

crates/burn-onnx/src/export/shape.rs
```diff
@@ -0,0 +1,726 @@
...
+                .ok_or(ExportError::MissingValue(validation_id))?;
+            for (axis, (&sample_dim, &validation_dim)) in sample_output
+                .shape
+                .iter()
+                .zip(validation_output.shape.iter())
+                .enumerate()
+            {
+                if sample_dim == validation_dim {
+                    continue;
+                }
```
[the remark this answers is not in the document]
antimora: Agreed, and it generalizes further than the example: after a stride-2 pool every consecutive pair `(2k, 2k+1)` collides, so 4/5, 8/9 and 16/17 all do this. Picking two nearby sizes for the sample and validation inputs is the natural thing to do.

The resulting artifact is the nasty part. The model advertises `dim_param: "height"` on its input while hard-coding the reshape, so it loads cleanly, runs correctly on the two sampled sizes, and only breaks at a third one.

crates/burn-onnx/src/export/shape.rs
```diff
@@ -0,0 +1,726 @@
...
+            let mut unresolved = Vec::new();
+            for (axis, (&sample_dim, &validation_dim)) in output
+                .shape
+                .iter()
+                .zip(validation_output.shape.iter())
+                .enumerate()
+            {
+                if sample_dim == validation_dim {
+                    dimensions.push(ShapeExpr::Static(sample_dim));
+                    continue;
```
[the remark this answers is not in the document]
antimora: Same root cause as the output-axis one, and this is the half that actually writes the wrong bytes: the `Reshape` shape initializer comes out as a constant `[1, 4]`.

Related, while you're in here: `Interpolate.output_size` and the constant-pad `pads` are also read straight from the sample trace rather than from `graph.shapes`, so the "lowering is intentionally independent from trace-based shape inference" note at the top of `lower/mod.rs` doesn't fully hold yet.

Cargo.toml
```diff
@@ -86,27 +86,32 @@ serde = { version = "1.0.228", default-features = false, features = [
...
+burn = { git = "https://github.com/tracel-ai/burn", default-features = false, rev = "e3b258cee0e314cd489177df42fa2d84ba450c46" }
+burn-flex = { git = "https://github.com/tracel-ai/burn", default-features = false, rev = "e3b258cee0e314cd489177df42fa2d84ba450c46" }
+burn-pack = { git = "https://github.com/tracel-ai/burn", default-features = false, rev = "e3b258cee0e314cd489177df42fa2d84ba450c46" }
+burn-store = { git = "https://github.com/tracel-ai/burn", default-features = false, rev = "e3b258cee0e314cd489177df42fa2d84ba450c46" }
+burn-tensor = { git = "https://github.com/tracel-ai/burn", default-features = false, rev = "e3b258cee0e314cd489177df42fa2d84ba450c46" }
 
 ### For local development. ###
 # burn = { path = "../burn/crates/burn", default-features = false }
 # burn-flex = { path = "../burn/crates/burn-flex", default-features = false }
+# burn-pack = { path = "../burn/crates/burn-flex", default-features = false }
```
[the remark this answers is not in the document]
antimora: Confirmed, `burn-pack` points at `../burn/crates/burn-flex`. The whole local-development block is new in this PR, so it's in scope to fix here.

antimora:
I have rebased so CI can pass now

laggui:
@antimora Thanks for the detailed review. I addressed the actionable comments.

I demoted the lower-level APIs to crate-only, as users are expected to use the supported high-level API as you pointed out. But `ExportInput` and `ExportOutput` (renamed) remain public, doc-hidden sealed traits because they appear in the bounds of the public `OnnxExporter` methods. They can't be implemented by downstream users, they simply define which types can cross the model's forward boundary.

I didn't reject distinct dynamic symbols merely because they have the same sample/validation pair. Square inputs commonly have independent "height" and "width" symbols with identical captured values, so that would reject valid exports. Instead, the dynamic-axis analysis now tracks symbol provenance through operations. Exact dependencies preserve the corresponding input symbol, while derived or ambiguous dimensions receive a fresh output symbol.

I also updated the burn revision, it was previously linked to unmerged changes that just landed on main. Had to update the `TensorSnapshot` usage given the changes in https://github.com/tracel-ai/burn/pull/5349

antimora (approved):
LGTM

PR #498, opened 2026-09-01
feat(export): support Conv1d, slicing, indexing, creation and reductions

Marc-AnthonyG:
Extends the ONNX exporter with support for:

- `slice` -> ONNX `Slice`, including negative steps
- `select` -> ONNX `Gather`
- `scatter` -> ONNX `ScatterElements`, including assign/add/mul/min/max reductions
- `zeros` / `ones` -> ONNX `ConstantOfShape` with explicit dtype casting
- `sum_dim` -> ONNX `ReduceSum`
- `argmax` / `argmin` -> ONNX `ArgMax` / `ArgMin`
- `max_dim_with_indices` / `min_dim_with_indices` -> `ArgMax` / `ArgMin` with `GatherElements`
- `conv1d` -> ONNX `Conv`

### DIFF
--- a/crates/burn-onnx/src/export/lower/base.rs
+++ b/crates/burn-onnx/src/export/lower/base.rs
@@ -3,13 +3,15 @@
 //! This family contains operations requiring ONNX-specific operands or
 //! attributes, including reshape shape expressions and concatenation axes.
 
-use burn::backend::ir::{BaseOperationIr, CreationOpIr, OperationIr, ScalarIr, SliceOpIr};
+use burn::backend::DType;
+use burn::backend::ir::{
+    BaseOperationIr, CreationOpIr, OperationIr, ScalarIr, SliceOpIr, TensorIr,
+};
 use burn::backend::tensor::IndexingUpdateOp;
 
 use crate::export::ExportError;
 
-use super::context::LoweringContext;
-use super::scalar_tensor;
+use super::{context::LoweringContext, onnx_dtype_parts, scalar_tensor};
 
 pub(super) fn lower(
     context: &mut LoweringContext<'_>,
@@ -53,7 +55,7 @@ pub(super) fn lower(
         BaseOperationIr::Select(select) => {
             let inputs = vec![
                 context.tensor_name(select.tensor.id),
-                context.tensor_name(select.indices.id),
+                lower_indices(context, index, "select", &select.indices)?,
             ];
             let output = context.tensor_name(select.out.id);
             context.node(format!("node_{index}"), "Gather", inputs, vec![output]);
@@ -63,7 +65,7 @@ pub(super) fn lower(
         BaseOperationIr::Scatter(scatter) => {
             let inputs = vec![
                 context.tensor_name(scatter.tensor.id),
-                context.tensor_name(scatter.indices.id),
+                lower_indices(context, index, "scatter", &scatter.indices)?,
                 context.tensor_name(scatter.value.id),
             ];
             let output = context.tensor_name(scatter.out.id);
@@ -87,34 +89,46 @@ pub(super) fn lower(
             Ok(true)
         }
         BaseOperationIr::Zeros(creation) => {
-            lower_creation(context, index, creation, 0.0)?;
+            lower_creation(context, index, creation, 0)?;
             Ok(true)
         }
         BaseOperationIr::Ones(creation) => {
-            lower_creation(context, index, creation, 1.0)?;
+            lower_creation(context, index, creation, 1)?;
             Ok(true)
         }
         _ => Ok(false),
     }
 }
 
+fn lower_indices(
+    context: &mut LoweringContext<'_>,
+    index: usize,
+    kind: &'static str,
+    indices: &TensorIr,
+) -> Result<String, ExportError> {
+    let input = context.tensor_name(indices.id);
+    if matches!(indices.dtype, DType::I32 | DType::I64) {
+        return Ok(input);
+    }
+    if !indices.dtype.is_int() && !indices.dtype.is_uint() {
+        return Err(ExportError::UnsupportedOperation {
+            operation: index,
+            kind: format!("{kind} with non-integer {:?} indices", indices.dtype),
+        });
+    }
+
+    let output = format!("node_{index}_{kind}_indices64");
+    context.node(output.clone(), "Cast", vec![input], vec![output.clone()]);
+    context.int_attribute("to", 7);
+    Ok(output)
+}
+
 fn lower_slice(context: &mut LoweringContext<'_>, index: usize, slice: &SliceOpIr) {
-    let starts: Vec<i64> = slice
+    let (starts, ends): (Vec<_>, Vec<_>) = slice
         .ranges
         .iter()
-        .map(|range| range.start as i64)
-        .collect();
-    let ends: Vec<i64> = slice
-        .ranges
-        .iter()
-        .map(|range| match range.end {
-            Some(end) => end as i64,
-            // An open end takes everything the step direction can reach; ONNX
-            // clamps out-of-range bounds to the dimension.
-            None if range.step >= 0 => i64::MAX,
-            None => i64::MIN,
-        })
-        .collect();
+        .map(|range| onnx_slice_bounds(range.start, range.end, range.step))
+        .unzip();
     let axes: Vec<i64> = (0..slice.ranges.len() as i64).collect();
     let steps: Vec<i64> = slice.ranges.iter().map(|range| range.step as i64).collect();
 
@@ -137,27 +151,57 @@ fn lower_slice(context: &mut LoweringContext<'_>, index: usize, slice: &SliceOpI
     );
 }
 
+fn onnx_slice_bounds(start: isize, end: Option<isize>, step: isize) -> (i64, i64) {
+    if step > 0 {
+        return (start as i64, end.map_or(i64::MAX, |end| end as i64));
+    }
+
+    // Burn selects an ascending [start, end) interval and traverses it in
+    // reverse. ONNX uses Python-style descending bounds, so swap the interval
+    // endpoints and make each one inclusive/exclusive in the other direction.
+    let onnx_start = end.map_or(-1, |end| end.saturating_sub(1) as i64);
+    let onnx_end = if start == 0 {
+        i64::MIN
+    } else {
+        start.saturating_sub(1) as i64
+    };
+    (onnx_start, onnx_end)
+}
+
 fn lower_creation(
     context: &mut LoweringContext<'_>,
     index: usize,
     creation: &CreationOpIr,
-    value: f64,
+    value: i64,
 ) -> Result<(), ExportError> {
     let shape_name = context.shape_input(index, creation.out.id)?;
     let output = context.tensor_name(creation.out.id);
+    let constant_output = if creation.out.dtype == DType::I64 {
+        output.clone()
+    } else {
+        format!("node_{index}_constant")
+    };
     context.node(
-        format!("node_{index}"),
+        format!("node_{index}_constant"),
         "ConstantOfShape",
         vec![shape_name],
-        vec![output],
+        vec![constant_output.clone()],
     );
     context.tensor_attribute(
         "value",
-        scalar_tensor(
-            creation.out.dtype,
-            ScalarIr::new(value, &creation.out.dtype),
-            creation.out.id,
-        )?,
+        scalar_tensor(DType::I64, ScalarIr::Int(value), creation.out.id)?,
     );
+    if constant_output != output {
+        context.node(
+            format!("node_{index}"),
+            "Cast",
+            vec![constant_output],
+            vec![output],
+        );
+        context.int_attribute(
+            "to",
+            onnx_dtype_parts(creation.out.id, creation.out.dtype)? as i64,
+        );
+    }
     Ok(())
 }
--- a/crates/burn-onnx/src/export/lower/base.rs
+++ b/crates/burn-onnx/src/export/lower/base.rs
@@ -119,7 +119,7 @@ fn lower_indices(
 
     let output = format!("node_{index}_{kind}_indices64");
     context.node(output.clone(), "Cast", vec![input], vec![output.clone()]);
-    context.int_attribute("to", 7);
+    context.int_attribute("to", onnx_dtype_parts(indices.id, DType::I64)? as i64);
     Ok(output)
 }
 
--- a/crates/burn-onnx/src/export/lower/base.rs
+++ b/crates/burn-onnx/src/export/lower/base.rs
@@ -155,6 +155,11 @@ fn onnx_slice_bounds(start: isize, end: Option<isize>, step: isize) -> (i64, i64
     if step > 0 {
         return (start as i64, end.map_or(i64::MAX, |end| end as i64));
     }
+    if end.is_some_and(|end| start >= end) {
+        // Burn permits empty intervals and returns no elements regardless of
+        // the step. Equal in-range ONNX bounds preserve that cardinality.
+        return (0, 0);
+    }
 
     // Burn selects an ascending [start, end) interval and traverses it in
     // reverse. ONNX uses Python-style descending bounds, so swap the interval
@@ -204,4 +209,15 @@ fn lower_creation(
         );
     }
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::onnx_slice_bounds;
+
+    #[test]
+    fn negative_step_keeps_empty_intervals_empty() {
+        assert_eq!(onnx_slice_bounds(0, Some(0), -1), (0, 0));
+        assert_eq!(onnx_slice_bounds(3, Some(1), -2), (0, 0));
+    }
 }
--- a/crates/burn-onnx/src/export/lower/base.rs
+++ b/crates/burn-onnx/src/export/lower/base.rs
@@ -3,11 +3,13 @@
 //! This family contains operations requiring ONNX-specific operands or
 //! attributes, including reshape shape expressions and concatenation axes.
 
-use burn::backend::ir::{BaseOperationIr, OperationIr};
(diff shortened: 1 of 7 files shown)


antimora (changes requested):
Please update the PR description; currently it's not even clear what are accomplishing.

Marc-AnthonyG:
> Please update the PR description; currently it's not even clear what are accomplishing. 

@laggui will look at it. I want to add some operations to the exporter:  Conv1d, Slice, Select, Scatter, MaxDimWithIndices, MinDimWithIndices.

It is just a slop starting point right now.

laggui (commented):
The follow-up changes complete the exporter contract that was missing from the original implementation:

- Translate Burn's negative-step slice ranges to ONNX semantics
- Cast `Gather` and `ScatterElements` indices to an ONNX-supported dtype
- Enforce opset 18 dtype constraints for convolution and `ReduceSum`
- Support additional dtypes for zeros/ones through `ConstantOfShape` + explicit cast
- Propagate dynamic axes

with added coverage.

These changes should also address the review comments.

crates/burn-onnx/src/export/shape/dynamic_axes.rs
```diff
@@ -260,6 +284,42 @@ impl AxisTracker {
...
+                            && sample_range.step.unsigned_abs() == 1
+                            && slice_covers_axis(sample_range, sample_dimension)
+                            && slice_covers_axis(validation_range, validation_dimension)
+                    },
+                );
+            if exact {
+                self.axis(&sample.tensor, axis, &sample.out, axis);
+            } else {
+                self.derived_axis(&sample.tensor, axis, &sample.out, axis);
+            }
```
antimora, line 319: This branch is where the runtime mismatch comes from. `into_slices` rewrites every range through `to_range(dim)` before capture, so `..` on an axis of size 2 arrives here as `0..2`. When the two captures collide on a derived axis the ranges are identical, the validator passes, `exact` propagates the dynamic marking to the output, and `lower_slice` still emits the captured `ends = 2`.

Repro with the same conv as `dynamic_conv1d_handles_colliding_capture_shapes`:

```rust
fn forward(&self, x: Tensor<3>) -> Tensor<3> {
    self.conv.forward(x).slice(s![.., 0..1, ..])
}
```

Sample `[1, 1, 4]`, validation `[1, 1, 5]`, length axis dynamic. Export succeeds and the output declares the length axis as a dim_param, but at runtime length 6 Burn returns `[1, 1, 3]` and ORT returns `[1, 1, 2]`. Same outcome for `s![.., .., ..;-1]`, `s![.., .., 1..]`, and `s![.., .., ..-1]`.

Two ways out: return `DynamicShapeLost` when a sliced axis is potentially dynamic (mirrors the existing varying-bounds rejection), or have the resolver emit open bounds (`i64::MAX` / `i64::MIN`) for full-coverage ranges and reject partial ones. The second also lifts the current restriction where any slice on a tensor with a dynamic batch axis fails, even `s![.., 1..3]`. Either way this needs a test that composes a shape-collapsing op with a slice; the differing-capture case is already covered and passes.

crates/burn-onnx/src/export/lower/base.rs
```diff
@@ -44,6 +48,176 @@ pub(super) fn lower(
...
+    // reverse. ONNX uses Python-style descending bounds, so swap the interval
+    // endpoints and make each one inclusive/exclusive in the other direction.
+    let onnx_start = end.map_or(-1, |end| end.saturating_sub(1) as i64);
+    let onnx_end = if start == 0 {
+        i64::MIN
+    } else {
+        start.saturating_sub(1) as i64
+    };
+    (onnx_start, onnx_end)
+}
```
antimora, line 174: The empty-interval fix from the last round is in, but it cannot be reached: `Tensor::slice` returns `Tensor::empty` before the backend records anything when an output dim is 0, so the IR never carries an empty Slice. More generally, `into_slices` normalizes every range before capture: `start >= 0`, `end` is always `Some` and clamped to the dim, and there is one range per axis. For that input the conversion is right. I checked `..;-1`, `-3..;-1`, `1..-1`, `0..4;-2`, `1..;-3`, `5..;-2`, and a two-axis reversal against ORT.

The signature and comments promise more than that though. With raw values it goes wrong in both directions: `(-1, Some(0), -1)` gives `(-1, -2)`, which ONNX reads as the last element, where Burn's normalized `[4, 0)` is empty; `(1, Some(-1), -1)` trips the `start >= end` guard and emits an empty slice where Burn gives `3, 2, 1`. Nothing produces those today, so either narrow the signature to the normalized form (`end: isize`, non-negative) and say so in a doc comment, or normalize against `slice.tensor.shape`, which is in scope at the call site.

Also worth a unit case for `start == 0` with a negative step. `s![..;-1]` is the most common reverse form and the `i64::MIN` mapping is the one thing here that is easy to get wrong.

crates/burn-onnx/src/export/lower/base.rs
```diff
@@ -44,6 +48,176 @@ pub(super) fn lower(
...
+
+fn lower_creation(
+    context: &mut LoweringContext<'_>,
+    index: usize,
+    creation: &CreationOpIr,
+    value: i64,
+) -> Result<(), ExportError> {
+    let shape_name = context.shape_input(index, creation.out.id)?;
+    let output = context.tensor_name(creation.out.id);
+    let constant_output = if creation.out.dtype == DType::I64 {
```
antimora, line 184: The int64 constant plus `Cast` handles every dtype now, which resolves the earlier concern, but it also puts a `Cast` on the common F32 path. `ConstantOfShape` takes F32/F64/I32/I64 values directly and `scalar_tensor` already covers those four; `Full` right next to this emits a single node. Consider emitting the native value for those and keeping int64 plus `Cast` as the fallback for F16/BF16/Bool/U8. Not a blocker.

crates/burn-onnx/src/export/lower/module.rs
```diff
@@ -19,7 +20,39 @@ pub(super) fn lower(
         return Ok(false);
     };
     match operation {
+        ModuleOperationIr::Conv1d(conv) => {
```
antimora, line 23: Not introduced here, Conv2d on main has the same issue, but Conv1d is new and its colliding-shapes test does not cover it. `PaddingConfig1d::Same` with stride > 1 resolves to different padding per input parity before capture: with kernel 3 and stride 2, odd lengths get symmetric `(1, 1)` and even lengths get a `Pad(0, 1)` op followed by a zero-padding conv. Sample 4 and validation 6 both capture the even form, export succeeds, and runtime length 5 diverges: Burn gives `[1, 1, 3]`, ORT gives `[1, 1, 2]`.

Burn does not keep the `Same` intent in the IR, so there is no clean lowering to `auto_pad`. A note in the export docs, or rejecting `Same` with stride > 1 under `export_dynamic`, would at least make it visible.

crates/burn-onnx/src/export/shape/mod.rs
```diff
@@ -640,6 +640,15 @@ fn shape_operations(graph: &GraphIr) -> impl Iterator<Item = ShapeOperation<'_>>
...
             }),
+            OperationIr::BaseFloat(BaseOperationIr::Zeros(op) | BaseOperationIr::Ones(op))
+            | OperationIr::BaseInt(BaseOperationIr::Zeros(op) | BaseOperationIr::Ones(op))
+            | OperationIr::BaseBool(BaseOperationIr::Zeros(op) | BaseOperationIr::Ones(op)) => {
+                Some(ShapeOperation {
+                    index,
+                    source: None,
+                    output: &op.out,
+                })
+            }
```
antimora, line 651: With Zeros and Ones registered as `source: None`, the two `DynamicShapeLost` messages that key on `source.is_none()` still say "Full" (lines 297 and 402 in this file). Someone hitting this from `Tensor::zeros` will be looking for an op that is not in their model. "created tensor dimension" or passing the op name through would fix it.

crates/burn-onnx/src/export/lower/numeric.rs
```diff
@@ -64,6 +118,112 @@ pub(super) fn lower(
...
+        reduce.axis,
+        arg_op,
+    )?;
+    Ok(())
+}
+
+/// Lower a max or min reduction that also yields the winning indices.
+///
+/// ONNX has no single operator for this pair. `ArgMax`/`ArgMin` produces the
+/// indices, and `GatherElements` reads the values back out through them.
```
antimora, line 142: "ONNX has no single operator for this pair" is not quite right. `TopK` with `k = 1` returns values and indices in one node, and its tie rule (lower index first) matches Burn. The two-node form is a reasonable choice since it shares the index path with plain argmax/argmin, but the comment should present it as a choice rather than a constraint, or the next person will not consider `TopK` when they need it.

crates/burn-onnx/src/export/lower/numeric.rs
```diff
@@ -64,6 +118,112 @@ pub(super) fn lower(
...
+    arg_op: &'static str,
+) -> Result<(), ExportError> {
+    // `GatherElements` needs indices of the input's rank, which only the
+    // dimension-keeping form provides.
+    if reduce.out.shape.num_dims() != reduce.tensor.shape.num_dims() {
+        return Err(ExportError::UnsupportedOperation {
+            operation: index,
+            kind: format!("{arg_op} reduction that drops the reduced dimension"),
+        });
+    }
```
antimora, line 156: On the earlier rank comment: `ReduceDimWithIndicesOpIr::create` sets `shape[dim] = 1` on both outputs, so the rank always matches and this branch cannot fire. Keeping it as a guard is fine, but the message reads like a user scenario and there is no Burn API that produces one. I would word it as an internal invariant.

Related: `lower_arg_indices` hardcodes `keepdims = 1` while the `SumDim` path computes it from the ranks. Both are constant for the same reason, so pick one style.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -471,6 +601,352 @@ fn small_cnn_matches_burn() {
...
+    assert!(
+        model.graph.output[0]
+            .type_
+            .as_ref()
+            .unwrap()
+            .tensor_type()
+            .shape
+            .dim[2]
+            .has_dim_param()
+    );
```
antimora, line 666: `has_dim_param()` also passes if the output wrongly inherits the input symbol (say the conv length axis went through `axis` instead of `derived_axis`), since "length" is a dim_param too, and ORT only warns on symbolic mismatches at run time. Asserting the param is not "length" (or equals the derived `output_0_dim_2`) pins what this test is actually about.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -471,6 +601,352 @@ fn small_cnn_matches_burn() {
...
+        shape.iter().map(|&dim| dim as usize).collect::<Vec<_>>(),
+    )
+    .assert_approx_eq::<f32>(&expected_negative.into_data(), Tolerance::default());
+}
+
+#[test]
+fn dynamic_slice_rejects_ambiguous_clamped_bounds() {
+    let device = Device::default();
+    // Burn clamps `..5` to each captured input. The resulting `0..3` and
+    // `0..4` ranges cannot be distinguished from a genuinely open range.
```
antimora, line 706: The comment implies an open range would be accepted, but it is not. The validator rejects any Slice whose recorded ranges differ between captures, and since `into_slices` bakes the dim into every axis, that is every slice on a tensor with a dynamic axis, including `s![.., 1..3]` with a dynamic batch. Worth saying plainly here and in the exporter docs, since it is the main limitation a user will hit. The open-bounds approach I suggested on `transfer_slice` would lift it.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -471,6 +601,352 @@ fn small_cnn_matches_burn() {
...
+    let input = Tensor::from_floats([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]], &device);
+    let input_values = input.clone().into_data().try_to_vec::<f32>().unwrap();
+    let expected = [
+        [[10.0f32, 2.0, -2.0], [20.0, 3.0, 6.0]],
+        [[11.0, 2.0, 1.0], [24.0, 8.0, 6.0]],
+        [[10.0, 2.0, -6.0], [80.0, 15.0, 6.0]],
+        [[1.0, 2.0, -2.0], [4.0, 3.0, 6.0]],
+        [[10.0, 2.0, 3.0], [20.0, 5.0, 6.0]],
+    ]
+    .map(TensorData::from);
```
antimora, line 760: The hand-computed rows are right (I recomputed all five), but this is the only `*_matches_burn` test that never calls `module.forward`, so it checks ORT against ONNX semantics rather than against Burn. If Burn's scatter reduction semantics drift on an upgrade, nothing here notices. Comparing to `module.forward(input)` like the others, with the literal array kept as a sanity anchor, closes that.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -471,6 +601,352 @@ fn small_cnn_matches_burn() {
...
+        .commit_from_memory(model.as_bytes())
+        .unwrap();
+    let outputs = session
+        .run(ort::inputs![
+            OrtTensor::from_array(([2, 4], input_values)).unwrap()
+        ])
+        .unwrap();
+    for (output, expected) in [(&outputs[0], expected_max), (&outputs[1], expected_min)] {
+        let (_, values) = output.try_extract_tensor::<i64>().unwrap();
+        TensorData::new(values.to_vec(), vec![2, 1]).assert_eq(&expected.into_data(), false);
```
antimora, line 917: `vec![2, 1]` hardcodes the shape, so if `keepdims` were dropped ORT would return `[2]` with the same values and this still passes. `dim_reductions_match_burn` uses the shape ORT returns; same thing here.

antimora (changes requested):
Ran the export unit tests and the 37 ORT tests locally, all green, and clippy is clean on the export paths. The description reads well now, thanks. The earlier Copilot points are addressed in the follow-up commits; where I comment on the same spots it is because the fix is in but the assumption around it still needs tightening.

One blocker: a slice after a shape-collapsing op on a dynamic axis exports successfully but truncates at runtime. Details inline on `transfer_slice`, with a repro that reuses the conv from the colliding-shapes test.

The rest is smaller: two error messages that still say "Full", a doc comment that rules out `TopK` when it should not, a `Same` padding hazard that Conv1d inherits from Conv2d, and three tests that would keep passing if the thing they cover regressed.

laggui:
@antimora sorry I merged this thinking everything had been addressed :sweat_smile: I'll open follow-up PR in the morning for your comments!

crates/burn-onnx/src/export/shape/dynamic_axes.rs
laggui: Addressed conservatively by returning `DynamicShapeLost` whenever a sliced axis is potentially dynamic. The normalized burn IR cannot distinguish an open bound from an explicit bound that was clamped during capture, so lifting apparently full-coverage ranges would be unsound.

crates/burn-onnx/src/export/lower/base.rs
laggui: Updated the docs to state that this helper operates on ranges already normalized by burn.

I didn't removed the unreachable empty-interval branch and its unit tests since empty slices, even if currently unreachable. I believe this should actually be fixed on burn's side. An empty slice operation should still emit a slice op IR. The `Tensor::empty` bypass should not happen at the high-level, but backend level.

crates/burn-onnx/src/export/lower/base.rs
laggui: Whoops, good catch. I updated the condition to match the supported dtypes for opset 18 (which used version 9 of `ConstantOfShape`).

crates/burn-onnx/src/export/lower/module.rs
laggui: I updated the burn revision which now supports asymmetric padding for conv1d and conv2d. Theior IR now carry explicit start/end padding pairs, and the xporter emits asymmetric ONNX pads directly instead of relying on a `Pad` operation.

For dynamic export, an asymmetric padded convolution with stride greater than one is rejected when its spatial axis is potentially dynamic.

The original `Same` intent is still not retained in burn, which resolves the explicit padding, so it remains indistinguishable from genuinely explicit padding.

crates/burn-onnx/src/export/shape/mod.rs
laggui: Updated both messages to refer to a "created tensor dimension", so it correctly applies to full/zeros/ones.

crates/burn-onnx/src/export/lower/numeric.rs
laggui: Reworded the comment.

crates/burn-onnx/src/export/lower/numeric.rs
laggui: Updated.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Strengthened the assertion to require the derived symbol `output_0_dim_2`.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Updated the test comment and public export_dynamic documentation to say plainly that slices over potentially dynamic axes are currently rejected. Burn normalizes every range against the captured dimension, so the exporter cannot safely distinguish open and explicitly clamped bounds.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Fixed.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Updated the test to construct `TensorData` using the shape returned by ORT

PR #502, opened 2026-09-03
fix(export): correct Conv, ArgMax, ArgMin, and ConstantOfShape lowering

laggui:
### Checklist

- [ ] Confirm relevant documentation has been updated.
- [ ] For new/modified ONNX operators: verified implementation matches `onnx-spec/ops/<OpName>.md`.

### Related Issues/PRs

Follow-up to #498
Addresses review comments https://github.com/tracel-ai/burn-onnx/pull/498#pullrequestreview-5094183178

### Changes

- Update burn and correctly lower asymmetric convolution padding (https://github.com/tracel-ai/burn/pull/5521)
- Preserve `ArgMax` / `ArgMin` output ranks
- Lower max/min with indices to `TopK` instead of `ArgMax`/`ArgMin` + `GatherElements`
- Emit opser 18 `ConstantOfShape` dtypes directly when supported
- Reject dynamics slices and convolution padding when capture loses runtime semantics

### Testing

Added coverage for these cases

### DIFF
--- a/crates/burn-onnx/src/export/exporter.rs
+++ b/crates/burn-onnx/src/export/exporter.rs
@@ -280,6 +280,11 @@ impl OnnxExporter {
     /// `input_specs` are positional and must contain one entry per tensor in
     /// `sample_inputs`. Static axes must agree between both input sets; dynamic
     /// axes must differ. Repeated symbols must refer to identical dimensions.
+    ///
+    /// Export fails when capture has normalized away shape-dependent intent,
+    /// including slices over potentially dynamic axes and input-dependent
+    /// padding before a strided convolution. Rejecting these cases prevents a
+    /// model that is valid only for the two captured shapes from being emitted.
     pub fn export_dynamic<M, I, O, F>(
         &self,
         module: &M,
--- a/crates/burn-onnx/src/export/exporter.rs
+++ b/crates/burn-onnx/src/export/exporter.rs
@@ -281,10 +281,12 @@ impl OnnxExporter {
     /// `sample_inputs`. Static axes must agree between both input sets; dynamic
     /// axes must differ. Repeated symbols must refer to identical dimensions.
     ///
-    /// Export fails when capture has normalized away shape-dependent intent,
-    /// including slices over potentially dynamic axes and input-dependent
-    /// padding before a strided convolution. Rejecting these cases prevents a
-    /// model that is valid only for the two captured shapes from being emitted.
+    /// Export fails when capture has normalized away shape-dependent intent.
+    /// Any slice of a tensor with a potentially dynamic axis is rejected,
+    /// including a full `..` range on that axis, because Burn clamps every
+    /// range to the captured shape. Input-dependent padding before a strided
+    /// convolution is rejected too. These checks are conservative and can
+    /// reject explicit padding that would export correctly.
     pub fn export_dynamic<M, I, O, F>(
         &self,
         module: &M,
--- a/crates/burn-onnx/src/export/lower/base.rs
+++ b/crates/burn-onnx/src/export/lower/base.rs
@@ -155,9 +155,10 @@ fn lower_slice(context: &mut LoweringContext<'_>, index: usize, slice: &SliceOpI
 ///
 /// At capture time Burn has already clamped the ascending interval to the
 /// input shape. A negative step reverses that normalized interval; it is not a
-/// raw Python-style descending slice. The empty-interval conversion is kept so
-/// capture remains correct once Burn moves its empty-output bypass from the
-/// Tensor API to execution backends and preserves Slice in the captured IR.
+/// raw Python-style descending slice. `Tensor::slice` currently returns an
+/// empty tensor before the backend operation is recorded, so an empty interval
+/// never reaches the captured IR. The branch keeps the conversion total in case
+/// that changes.
 fn onnx_slice_bounds(start: isize, end: Option<isize>, step: isize) -> (i64, i64) {
     if step > 0 {
         return (start as i64, end.map_or(i64::MAX, |end| end as i64));
@@ -186,7 +187,12 @@ fn lower_creation(
 ) -> Result<(), ExportError> {
     let shape_name = context.shape_input(index, creation.out.id)?;
     let output = context.tensor_name(creation.out.id);
-    let constant_dtype = constant_of_shape_dtype(creation.out.dtype);
+    let constant_dtype = constant_of_shape_dtype(creation.out.dtype).ok_or_else(|| {
+        ExportError::UnsupportedDType {
+            tensor: creation.out.id,
+            dtype: format!("{:?}", creation.out.dtype),
+        }
+    })?;
     let constant_output = if constant_dtype == creation.out.dtype {
         output.clone()
     } else {
@@ -223,9 +229,10 @@ fn lower_creation(
 
 /// Return the dtype used for the `ConstantOfShape` value attribute.
 ///
-/// Opset 18 uses the version 9 `ConstantOfShape` schema. Unsupported output
-/// types are created as `I64` and converted by the following `Cast` node.
-fn constant_of_shape_dtype(dtype: DType) -> DType {
+/// Opset 18 uses the version 9 `ConstantOfShape` schema. `BF16` outputs are
+/// created as `F32` and converted by the following `Cast` node. Burn-specific
+/// dtypes without an ONNX representation are rejected.
+fn constant_of_shape_dtype(dtype: DType) -> Option<DType> {
     match dtype {
         DType::F16
         | DType::F32
@@ -238,8 +245,9 @@ fn constant_of_shape_dtype(dtype: DType) -> DType {
         | DType::U16
         | DType::U32
         | DType::U64
-        | DType::Bool(_) => dtype,
-        _ => DType::I64,
+        | DType::Bool(_) => Some(dtype),
+        DType::BF16 => Some(DType::F32),
+        DType::Flex32 | DType::QFloat(_) => None,
     }
 }
 
@@ -267,9 +275,10 @@ mod tests {
         ];
 
         for dtype in native_dtypes {
-            assert_eq!(constant_of_shape_dtype(dtype), dtype);
+            assert_eq!(constant_of_shape_dtype(dtype), Some(dtype));
         }
-        assert_eq!(constant_of_shape_dtype(DType::BF16), DType::I64);
+        assert_eq!(constant_of_shape_dtype(DType::BF16), Some(DType::F32));
+        assert_eq!(constant_of_shape_dtype(DType::Flex32), None);
     }
 
     #[test]
--- a/crates/burn-onnx/src/export/lower/base.rs
+++ b/crates/burn-onnx/src/export/lower/base.rs
@@ -151,16 +151,21 @@ fn lower_slice(context: &mut LoweringContext<'_>, index: usize, slice: &SliceOpI
     );
 }
 
+/// Convert a range normalized by Burn's slice API into ONNX bounds.
+///
+/// At capture time Burn has already clamped the ascending interval to the
+/// input shape. A negative step reverses that normalized interval; it is not a
+/// raw Python-style descending slice. The empty-interval conversion is kept so
+/// capture remains correct once Burn moves its empty-output bypass from the
+/// Tensor API to execution backends and preserves Slice in the captured IR.
 fn onnx_slice_bounds(start: isize, end: Option<isize>, step: isize) -> (i64, i64) {
     if step > 0 {
         return (start as i64, end.map_or(i64::MAX, |end| end as i64));
     }
     if end.is_some_and(|end| start >= end) {
-        // Burn permits empty intervals and returns no elements regardless of
-        // the step. Equal in-range ONNX bounds preserve that cardinality.
+        // Equal in-range ONNX bounds preserve Burn's empty interval.
         return (0, 0);
     }
-
     // Burn selects an ascending [start, end) interval and traverses it in
     // reverse. ONNX uses Python-style descending bounds, so swap the interval
     // endpoints and make each one inclusive/exclusive in the other direction.
@@ -181,7 +186,8 @@ fn lower_creation(
 ) -> Result<(), ExportError> {
     let shape_name = context.shape_input(index, creation.out.id)?;
     let output = context.tensor_name(creation.out.id);
-    let constant_output = if creation.out.dtype == DType::I64 {
+    let constant_dtype = constant_of_shape_dtype(creation.out.dtype);
+    let constant_output = if constant_dtype == creation.out.dtype {
         output.clone()
     } else {
         format!("node_{index}_constant")
@@ -194,7 +200,11 @@ fn lower_creation(
     );
     context.tensor_attribute(
         "value",
-        scalar_tensor(DType::I64, ScalarIr::Int(value), creation.out.id)?,
+        scalar_tensor(
+            constant_dtype,
+            ScalarIr::new(value, &constant_dtype),
+            creation.out.id,
+        )?,
     );
     if constant_output != output {
         context.node(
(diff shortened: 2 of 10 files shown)


antimora:
Checking locally ...

antimora (commented):
Ran the export suite on the head (974 lib + 43 ORT tests pass) and tried a few extra Conv configurations against the new dynamic-export checks. The lowering changes look right: the pads order matches ONNX for 1D and 2D, the TopK attributes, output order and tie-break match TopK-11 and Burn's contract, and the ConstantOfShape dtype list matches version 9.

The one thing to fix before merge is the `Same` padding gap in `validate_strided_convolution_padding` (repro inline). The rest are test blind spots and comment corrections.

Two things outside this diff, for a follow-up: `MaxPool2d`/`AvgPool2d` with `Same` and stride > 1 have the same class of gap and the new check does not cover them, and `NumericOperationIr::Full` still calls `scalar_tensor` with the output dtype directly, so BF16 `full` errors while BF16 `zeros`/`ones` now take the I64 + Cast path.

crates/burn-onnx/src/export/shape/mod.rs
```diff
@@ -160,11 +160,87 @@ impl<'a> PairedTraceAnalysis<'a> {
...
+        // the original `Same` intent is not retained in Conv*OpIr. Symmetric
+        // `Same` remains indistinguishable from explicit padding; Burn IR must
+        // retain the original padding intent to close that gap without
+        // rejecting valid explicit-padding convolutions.
+        let spatial_offset = input.shape.num_dims().saturating_sub(D);
+        for (spatial_axis, &stride) in stride.iter().enumerate() {
+            let axis = spatial_offset + spatial_axis;
+            if stride > 1
+                && self.potentially_dynamic.contains(input.id, axis)
+                && padding[spatial_axis].0 != padding[spatial_axis].1
```
antimora: This only catches `Same` when the captured widths happen to be asymmetric. Burn's `Same` total is `k - r` with `r = len mod stride` (0 mapped to stride), so it depends only on the residue. Two captures with the same residue always agree, and when that shared padding is symmetric nothing rejects it.

Repro: `Conv1dConfig::new(1, 2, 3).with_stride(2).with_padding(PaddingConfig1d::Same)`, sample length 5, validation length 7. Both capture `(1, 1)`, export succeeds with `pads = [1, 1]`, ORT matches Burn at lengths 5 and 7 and diverges at 6 and 8, where Burn pads `(0, 1)`.

`GraphStructureValidator` already rejects captures whose padding differs, so the only hole is equal residues. Rejecting when `sample_len % stride == validation_len % stride` on a potentially dynamic spatial axis with stride > 1 closes it: with different residues `Same` padding always differs and gets caught, with equal residues nothing can be proven either way. That subsumes the asymmetric check, and it turns the permanent false positive on explicit asymmetric padding into something the user fixes by picking a different validation length. The validation trace is at `self.resolver.validation.operations[index]`. A symmetric sibling of `dynamic_padded_strided_conv1d_is_rejected` with lengths 5 and 7 would pin it.

crates/burn-onnx/src/export/shape/mod.rs
```diff
@@ -160,11 +160,87 @@ impl<'a> PairedTraceAnalysis<'a> {
...
+        input: &TensorIr,
+        stride: &[usize; D],
+        padding: &[(usize, usize); D],
+    ) -> Result<(), ExportError> {
+        // Burn resolves `Same` to concrete start/end widths using the captured
+        // input shape. Asymmetric widths can change with the runtime input but
+        // the original `Same` intent is not retained in Conv*OpIr. Symmetric
+        // `Same` remains indistinguishable from explicit padding; Burn IR must
+        // retain the original padding intent to close that gap without
+        // rejecting valid explicit-padding convolutions.
```
antimora: Worth stating here that padding which differs between the two captures is already rejected by `GraphStructureValidator` (`normalize_structure` does not erase conv options), which is why reading only the sample trace is fine. Otherwise the next reader adds a validation-trace comparison.

That path reports `non-shape attributes differ`, which hides that `Same` padding is the cause. A Conv arm in `validate_shape_sensitive_operations` reusing the reason string below would give one consistent error for both cases.

crates/burn-onnx/src/export/shape/mod.rs
```diff
@@ -160,11 +160,87 @@ impl<'a> PairedTraceAnalysis<'a> {
...
+                | OperationIr::BaseInt(BaseOperationIr::Slice(slice))
+                | OperationIr::BaseBool(BaseOperationIr::Slice(slice)) => {
+                    if let Some(axis) = (0..slice.ranges.len())
+                        .find(|&axis| self.potentially_dynamic.contains(slice.tensor.id, axis))
+                    {
+                        return Err(ExportError::DynamicShapeLost {
+                            tensor: slice.out.id,
+                            axis,
+                            reason: format!(
+                                "operation {index} has normalized slice bounds on a potentially dynamic axis"
```
antimora: When this fires on a dynamic batch axis the user never sliced, the message will not tell them why. Something like `operation {index} slices a tensor with a potentially dynamic axis; Burn clamps every range, including \`..\`, to the captured shape, so the intended bounds cannot be recovered` saves a round trip.

crates/burn-onnx/src/export/exporter.rs
```diff
@@ -280,6 +280,11 @@ impl OnnxExporter {
     /// `input_specs` are positional and must contain one entry per tensor in
     /// `sample_inputs`. Static axes must agree between both input sets; dynamic
     /// axes must differ. Repeated symbols must refer to identical dimensions.
+    ///
+    /// Export fails when capture has normalized away shape-dependent intent,
+    /// including slices over potentially dynamic axes and input-dependent
+    /// padding before a strided convolution. Rejecting these cases prevents a
+    /// model that is valid only for the two captured shapes from being emitted.
```
antimora: Both halves of this drift from what the checks do. Slices: any `slice` of a tensor with a potentially dynamic axis is rejected, including a plain `..` on a dynamic batch axis, because `into_slices` rewrites `..` to `0..n` before capture. `x.slice(s![.., 0..1, ..])` with a dynamic batch axis fails today. Padding: the check rejects explicit asymmetric padding with stride > 1 and lets symmetric `Same` through. Suggest:

```rust
/// Export fails when capture has normalized away shape-dependent intent. Any
/// `slice` of a tensor with a potentially dynamic axis is rejected, including a
/// full `..` range on that axis, because Burn clamps every range to the captured
/// shape. Asymmetric padding before a strided convolution on such an axis is
/// rejected too. These checks are conservative: they can reject explicit padding
/// that would export correctly and do not detect `Same` padding whose captured
/// widths are symmetric.
```

Drop the last clause if the residue check lands.

crates/burn-onnx/src/export/lower/base.rs
```diff
@@ -151,16 +151,21 @@ fn lower_slice(context: &mut LoweringContext<'_>, index: usize, slice: &SliceOpI
     );
 }
 
+/// Convert a range normalized by Burn's slice API into ONNX bounds.
+///
+/// At capture time Burn has already clamped the ascending interval to the
+/// input shape. A negative step reverses that normalized interval; it is not a
+/// raw Python-style descending slice. The empty-interval conversion is kept so
+/// capture remains correct once Burn moves its empty-output bypass from the
+/// Tensor API to execution backends and preserves Slice in the captured IR.
```
antimora: This predicts what Burn will do and reads as stale either way. Today `Tensor::slice` returns `empty` before `K::slice` is called, so an empty interval never reaches the IR. Suggest saying that instead: "`Tensor::slice` currently returns an empty tensor before the backend op is recorded, so an empty interval never reaches the captured IR. The branch keeps the conversion total in case that changes."

crates/burn-onnx/src/export/lower/numeric.rs
```diff
@@ -187,6 +223,7 @@ fn lower_arg_indices(
     output: &TensorIr,
     axis: usize,
     arg_op: &'static str,
+    keepdims: bool,
 ) -> Result<String, ExportError> {
```
antimora: `lower_arg` is the only caller now and drops the name, so the `String` return and the `indices64.clone()` in the node push can go. `Result<(), ExportError>` is enough.

crates/burn-onnx/src/export/lower/numeric.rs
```diff
@@ -125,53 +125,89 @@ fn lower_arg(
...
-        vec![output],
+        "TopK",
+        vec![input, k],
+        vec![output, indices64.clone()],
     );
     context.int_attribute("axis", reduce.dim as i64);
+    context.int_attribute("largest", largest as i64);
+    context.int_attribute("sorted", 1);
+
+    if indices64 != indices_output {
```
antimora, line 214: Capture defaults ints to I64, so this branch (and the matching one in `lower_arg_indices`) is never taken by any test. A hand-built `ReduceDimWithIndicesOpIr` with I32 indices through `export_graph_with_bindings_and_opset`, like the test in `lower/mod.rs`, would cover the output order and the Cast wiring.

crates/burn-onnx/src/export/lower/numeric.rs
```diff
@@ -125,53 +125,89 @@ fn lower_arg(
...
     reduce: &ReduceDimWithIndicesOpIr,
-    arg_op: &'static str,
+    largest: bool,
+    operation: &'static str,
 ) -> Result<(), ExportError> {
-    // `GatherElements` needs indices of the input's rank, which only the
-    // dimension-keeping form provides.
-    if reduce.out.shape.num_dims() != reduce.tensor.shape.num_dims() {
+    // Opset 18 selects TopK-11, which does not support BF16.
+    if reduce.tensor.dtype == DType::BF16 {
```
antimora: `SumDim` and the conv lowering use allowlists for opset 18 dtypes; this is a denylist, so any dtype `onnx_dtype_parts` learns to map later gets past it without a TopK-11 check. Matching the allowlist style keeps the three in sync.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -609,7 +670,7 @@ fn conv1d_matches_burn() {
             .with_stride(2)
             .with_dilation(2)
             .with_groups(2)
-            .with_padding(PaddingConfig1d::Explicit(2, 2))
+            .with_padding(PaddingConfig1d::Explicit(1, 2))
```
antimora, line 704: This still passes with the old `[left, left]` lowering. Kernel 3 with dilation 2 is effectively 5; at stride 2 and length 9, padding `(1, 1)` and `(1, 2)` give padded lengths 11 and 12, both of which fit exactly four windows at the same offsets, so the extra end pad is never read. Length 8 or 10 changes the window count; `Explicit(1, 3)` also works.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -623,6 +684,26 @@ fn conv1d_matches_burn() {
     actual.assert_approx_eq::<f32>(&expected, Tolerance::rel_abs(RTOL, ATOL));
 }
 
+#[test]
+fn asymmetric_conv2d_matches_burn() {
+    let device = Device::default();
+    let module = SmallConv2d {
+        conv: Conv2dConfig::new([1, 2], [3, 2])
+            .with_stride([2, 1])
+            .with_padding(PaddingConfig2d::Explicit(1, 0, 2, 1))
```
antimora, line 724: Same blind spot on the height axis: `(1, 2)` at stride 2 on height 5 gives the same output as `(1, 1)`. Only the width axis `(0, 1)` at stride 1 distinguishes begin from end here. Height 4 or 6 makes both axes sensitive.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -153,6 +179,28 @@ impl ScatterColumns {
         })
         .into()
     }
+
+    fn reference(&self, input: Tensor<2>) -> Vec<Tensor<2>> {
+        // The default test backend only implements element-wise scatter for
+        // `Add`. Its ScatterND implementation supports every reduction and is
+        // equivalent here because all update coordinates are unique.
```
antimora: burn-flex's `float_scatter` has `Assign`, `Add` and `Mul` arms; only `Min` and `Max` are missing. Suggest: "The default backend has no element-wise scatter for `Min` and `Max`. Its ScatterND implementation supports every reduction and is equivalent here because all update coordinates are unique."

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -697,13 +874,21 @@ fn stepped_slice_matches_burn() {
...
 
 #[test]
 fn dynamic_slice_rejects_ambiguous_clamped_bounds() {
     let device = Device::default();
-    // Burn clamps `..5` to each captured input. The resulting `0..3` and
-    // `0..4` ranges cannot be distinguished from a genuinely open range.
+    // Burn normalizes every range before capture. For example, it clamps `..5`
+    // to `0..3` and `0..4` here, but the exporter cannot distinguish those
+    // bounds from a genuinely open range. It therefore rejects every slice on
+    // a potentially dynamic axis, even if two normalized captures agree.
```
antimora: This test asserts on `varying slice bounds`, which comes from `validate_shape_sensitive_operations` and runs before the new check in `validate_dynamic_operation_contracts`. The last sentence describes the new check, which this test never reaches; `dynamic_slice_after_shape_collapse_is_rejected` is the one that covers it. Move that sentence there and end this comment with "The bounds differ between captures, so the shape-sensitive check rejects the slice before dynamic-axis analysis runs."

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -654,24 +735,120 @@ fn dynamic_conv1d_handles_colliding_capture_shapes() {
...
             .shape
             .dim[2]
-            .has_dim_param()
+            .dim_param(),
+        "output_0_dim_2"
+    );
+}
+
+#[test]
+fn dynamic_slice_after_shape_collapse_is_rejected() {
```
antimora, line 783: All three slice tests under `export_dynamic` are rejections, so an over-eager check (wrong tensor id, wrong axis set) would go unnoticed. One accepting case would pin it, e.g. slicing a second static input, or `input.sum_dim(0).slice(s![.., 0..2])` where the dynamic axis has been reduced away.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -654,24 +735,120 @@ fn dynamic_conv1d_handles_colliding_capture_shapes() {
...
+
+    assert!(matches!(
+        result,
+        Err(ExportError::DynamicShapeLost { reason, .. })
+            if reason.contains("normalized slice bounds")
+    ));
+}
+
+#[test]
+fn dynamic_padded_strided_conv1d_is_rejected() {
```
antimora, line 851: Only rejections here too. If the `stride > 1` guard were dropped, every asymmetric conv on a dynamic axis would be rejected and nothing would fail. A positive case covers it: explicit asymmetric padding at stride 1 on a dynamic length, exported and run through ORT at a length neither capture used. `Explicit(0, 1)` with kernel 2, captures 5 and 7, runtime 6 exports and matches today.

crates/burn-onnx/tests/ort_numerical.rs
```diff
@@ -818,6 +1023,26 @@ fn dynamic_bool_zeros_and_ones_match_burn() {
     assert!(ones.iter().all(|value| *value));
 }
 
+#[test]
+fn native_constant_of_shape_dtypes_do_not_add_casts() {
```
antimora, line 1196: This only checks node types, so a wrong `value` attribute for F32 would pass. Run ORT here like the bool test does and assert the zeros and ones. Two more gaps from the same change: the I64 + Cast fallback is no longer exercised at all now that Bool is native (`Tensor::zeros(shape, (&device, DType::BF16))` asserting `["ConstantOfShape", "Cast"]` with `to == 16` covers it without ORT), and no non-F32 native dtype is checked end to end (I32 or F16 through the same `(&device, dtype)` form).

crates/burn-onnx/src/export/shape/mod.rs
laggui: Good catch. I changed the check to reject when both captures' padding is compatible compatible with Burn's `Same` calculation on a potentially dynamic strided axis.

crates/burn-onnx/src/export/shape/mod.rs
laggui: Updated.

crates/burn-onnx/src/export/shape/mod.rs
laggui: Done.

crates/burn-onnx/src/export/exporter.rs
laggui: Updated.

crates/burn-onnx/src/export/lower/base.rs
laggui: Fixed.

crates/burn-onnx/src/export/lower/numeric.rs
laggui: Replaced with an allow list.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Good catch. I changed the input length to 8.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Changed the height to 4, making the asymmetric height padding observable as well as the width padding.

crates/burn-onnx/tests/ort_numerical.rs
laggui: Fixed the comment

crates/burn-onnx/tests/ort_numerical.rs
laggui: Added an accepting case that reduces the dynamic axis with `sum_dim(0)` before slicing

crates/burn-onnx/tests/ort_numerical.rs
laggui: Added an explicit asymmetric-padding convolution test

crates/burn-onnx/tests/ort_numerical.rs
laggui: Done. The F32 case now runs through ORT. I also added an I32 (native) and BF16 case (constant + cast).

antimora (approved):
Verified on caf2d0c5: the export suite passes (976 lib + 49 ORT), the `Same` kernel 3 / stride 2 case with captures 5 and 7 that used to export wrong is now rejected, and explicit padding with a differing residue exports and matches ORT at runtime lengths neither capture used.

One follow-up, not blocking: `Valid` is `(0, 0)`, and `Same` also yields `(0, 0)` when the kernel equals the stride and the length is a multiple of it. A patch-embedding conv with kernel 16, stride 16 and captures 224/256 is rejected with the padding-intent message even though it has no padding; 224/250 exports fine. Worth adding to the error text: choose a validation shape whose `Same` widths differ from the captured padding.

Issue #350, opened 2026-04-15
Misc small buckets: max/pow/Selu/Pad-runtime-axes/Loop-rank-0/Squeeze multi-axis (12 tests)


antimora:
## Summary

Small buckets that each block 2–3 tests; grouped here to avoid ticket inflation while still tracking them individually under #314.

| Count | Error | Example | Notes |
|---:|---|---|---|
| 3 | `max: unsupported input types` | `test_center_crop_pad_crop_and_pad_expanded` | `max.rs:30` panics for non-on-device inputs (Shape, scalar combos). Add explicit branches. |
| 2 | `pow function requires RHS to be int or float type` | `test_pow_types_float32_uint32` | `pow.rs:35` rejects u32/u64 RHS; widen to cover all int dtypes. |
| 2 | `Selu: custom alpha (2) is not supported by Burn` | `test_selu`, `test_selu_example` | Emit manual `gamma * (elu(x, alpha))` formula when alpha/gamma differ from ONNX defaults; burn has `elu` with custom alpha. |
| 2 | `Pad: runtime axes input is not supported` | `test_constant_pad_axes`, `test_constant_pad_negative_axes` | Codegen would need to read axes at runtime. Separate from runtime-pads (#TODO). |
| 2 | `Loop body: rank-0 identity` | `test_range_float_type_positive_delta_expanded` | Loop body type inference treats rank-0 tensor as real rather than Scalar. |
| 1 | runtime `Squeeze` multi-axis negative ordering | `test_squeeze_negative_axes` | Was flipped then reverted to `fail-compare`; burn's `squeeze_dims` needs investigation on multi-axis runtime inputs. |

## Refs

- #314 — parent scoreboard issue

BenFradet:
I'm working on the max issue 👍

antimora:
> I'm working on the max issue 👍

Sounds good; Just beware of upcoming release (1 week maybe?); @laggui is leading the release.

antimora:
Rechecked on main (fedcb0b7). max (thanks @BenFradet), pow with unsigned RHS, Selu with custom alpha, Pad runtime axes and multi-axis Squeeze all pass now. One item is left: the Loop-body rank-0 Identity, which fails with `Node 'identity2' output 'identity2_out1' has rank-0 tensor. Rank-0 tensors should be Scalar` on 4 tests: `test_range_{float,int32,float16,bfloat16}_type_*_expanded`. Keeping this open for that.

PR #503, opened 2026-09-03
fix: support scalar and shape types for max

BenFradet:
### Checklist

### Related Issues/PRs

- https://github.com/tracel-ai/burn-onnx/issues/350
- [sub](https://github.com/tracel-ai/burn-onnx/blob/main/crates/burn-onnx/src/import/burn/node/sub.rs) was used as a reference

### Changes

- add more types support for max: scalar and shapes
- I'll do `min` next since it suffers from the same shortcomings

### Testing

- new insta tests
- new integration tests


crates/onnx-official-tests/expectations.toml
```diff
@@ -3110,11 +3104,11 @@ reason = "burn-onnx does not currently support 3D max pooling: Burn lacks a nati
 
 [test_maxpool_with_argmax_2d_precomputed_pads]
 status = "skip-compile"
-reason = "cannot find value `maxpool2d1_out2` in this scope"
+reason = "cannot find value `maxpool2d1_out2` in this scope: not found in this scope"
```
BenFradet, line 3107: these are the results of the `cargo xtask retriage` and `cargo xtask update-expectations` workflow

crates/burn-onnx/src/import/burn/node/max.rs
```diff
@@ -27,7 +27,90 @@ impl NodeCodegen for onnx_ir::node::max::MaxNode {
...
+                    for (result_item, rhs_item) in result.iter_mut().zip(#rhs.iter()) {
+                        *result_item = (*result_item).max(*rhs_item);
+                    }
+                    result
+                }
+            },
+
+            (lhs_ty, ArgType::Shape(_)) if lhs_ty.is_scalar() => {
+                let scalar_expr = if lhs_ty.is_scalar_tensor() {
+                    on_device_to_native(lhs.clone(), &lhs_ty.elem_type())
```
antimora: `on_device_to_native` emits only `(x).into_scalar::<T>()` with no widening (`argument_helpers.rs:78-81`). When the ScalarTensor is not i64, the generated code does not compile. Probed with `DType::F32`:

```rust
let __scalar = (rhs).into_scalar::<f32>();
*result_item = (*result_item).max(__scalar);   // i64::max(f32): no such impl
```

The snapshot test still passes because snapshots compare token text and never compile the result. Every snapshot on this arm uses `DType::I64`, which is why it is invisible. Same applies to line 70.

The correct version already exists in `concat.rs:5-16`:

```rust
fn scalar_as_i64(arg: &Argument, scope: &mut ScopeAtPosition<'_>) -> TokenStream {
    let value = scope.arg(arg);
    match &arg.ty {
        ArgType::ScalarTensor(dtype) => {
            let native = on_device_to_native(value, dtype);
            quote! { #native as i64 }        // <- the missing cast
        }
        _ => quote! { #value as i64 },
    }
}
```

Suggest lifting `scalar_as_i64` into `argument_helpers.rs` and exporting it from `prelude.rs` (which already re-exports that module). `add.rs`, `sub.rs` and `div.rs` all carry the same buggy copy, so this removes a four-way duplicate of the same 17-line block at the same time.

crates/burn-onnx/src/import/burn/node/max.rs
```diff
@@ -27,7 +27,90 @@ impl NodeCodegen for onnx_ir::node::max::MaxNode {
...
+                quote! { #lhs.clamp_min(#rhs) }
+            }
+            (ArgType::ScalarNative(_), rhs_ty) if rhs_ty.is_on_device() => {
+                quote! { #rhs.clamp_min(#lhs) }
+            }
+
+            (ArgType::Shape(_), ArgType::Shape(_)) => quote! {
+                {
+                    let mut result = #lhs;
+                    for (result_item, rhs_item) in result.iter_mut().zip(#rhs.iter()) {
```
antimora, line 44: `zip` silently truncates when the two shape ranks differ. I built a model that `onnx.checker.check_model` accepts, `Max(Shape(rank-1 input), Shape(rank-4 input))`:

- ONNX `ReferenceEvaluator`: `[3, 30, 4, 5]`
- Generated: `pub fn forward(&self, input1: Tensor<1>, input2: Tensor<4>) -> [i64; 1]`, returns `[3]`

Wrong length, wrong values, no error. Root cause is upstream in `onnx-ir/src/processor.rs:739-748`: `broadcast_output_type` takes the rank of the *first* `Shape` input and discards the other, so this loop faithfully implements a wrong declared type.

Pre-existing, not introduced here (I reproduced the identical `[i64; 1]` output with `Add`). But it changes the failure mode for `Max` specifically: before this PR the combination hit `panic!("max: unsupported input types")` and failed loudly, now it joins Add/Sub/Div in returning a silently wrong answer. Either add a rank check here or file it against `broadcast_output_type` since it affects four operators. Fine either way, just do not let it disappear.

crates/burn-onnx/src/import/burn/node/max.rs
```diff
@@ -27,7 +27,90 @@ impl NodeCodegen for onnx_ir::node::max::MaxNode {
...
+            }
+
+            (lhs_ty, ArgType::Shape(_)) if lhs_ty.is_on_device() => {
+                let dtype_tokens = lhs_ty.elem_type().to_tokens();
+                let rhs_tensor = quote! {
+                    Tensor::<1, burn::tensor::Int>::from_data(
+                        burn::tensor::TensorData::from(&#rhs as &[i64]),
+                        (&self.device, burn::tensor::DType::I64),
+                    )
+                    .cast(#dtype_tokens)
```
antimora: This arm is unguarded for non-integer dtypes. In Burn, `impl Cast<D, Int> for DType` has `type OutputKind = Int` and its doc comment says it panics when the dtype is not an int variant (`burn-tensor/src/tensor/api/cast.rs:86-95`). So with `DType::F32` the expression stays `Tensor<_, Int>`, fails to typecheck against `max_pair(Tensor<_, Float>)`, and would panic if it did compile. Same for the mirrored arm at line 105.

Only reachable via non-conformant ONNX (Max constrains all inputs to one `T`, and a `Shape` operand is int64), so low priority. Worth a guard anyway: the code this replaces gave a clear codegen message for this input, and now it produces an opaque rustc error against generated source the user never wrote.

Minor, same lines: for the common i64 case this emits a no-op `.cast(burn::tensor::DType::I64)`. Visible in the generated `max_shape_tensor.rs`. Harmless, but easy to skip.

crates/onnx-tests/tests/max/max_shape.py
```diff
@@ -0,0 +1,119 @@
...
+    outputs = session.run(None, {"input1": test_input1, "input2": test_input2})
+
+    max_shape_scalar, max_scalar_shape, max_shape_shape = outputs
+
+    print(f"\nTest output max_shape_scalar: {repr(max_shape_scalar)}")
+    print(f"Test output max_scalar_shape: {repr(max_scalar_shape)}")
+    print(f"Test output max_shape_shape: {repr(max_shape_shape)}")
+
+    # Verify results are the same
+    assert np.array_equal(max_shape_scalar, max_scalar_shape), (
```
antimora, line 112: This only asserts that `Max` is commutative, comparing two outputs of the same evaluator. It passes regardless of what the model computes, so it gives no signal. `max_shape_shape` is never asserted on at all.

`max_scalar.py:80-82` gets this right by comparing against PyTorch as an independent oracle. Here there is no torch model, so just assert the hardcoded expected arrays:

```python
np.testing.assert_array_equal(max_shape_scalar, [10, 8, 7])
np.testing.assert_array_equal(max_scalar_shape, [10, 8, 7])
np.testing.assert_array_equal(max_shape_shape, [10, 30, 6])
```

Those are the same values the Rust test asserts, so the script becomes the thing that justifies them.

crates/onnx-tests/tests/max/max_shape_tensor.py
```diff
@@ -0,0 +1,91 @@
...
+    session = ReferenceEvaluator(onnx_name, verbose=0)
+    outputs = session.run(None, {"input_tensor": test_input, "input_1d": test_1d})
+
+    max_shape_tensor, max_tensor_shape = outputs
+
+    print(f"\nTest output max_shape_tensor: {repr(max_shape_tensor)}")
+    print(f"Test output max_tensor_shape: {repr(max_tensor_shape)}")
+
+    # Verify results are the same
+    assert np.array_equal(max_shape_tensor, max_tensor_shape), (
```
antimora, line 96: Same as in `max_shape.py`: comparing two outputs of the same evaluator only asserts commutativity. Replace with `np.testing.assert_array_equal(max_shape_tensor, [5, 30, 9])` and the same for `max_tensor_shape`.

Separately, this model pairs a `Shape(3)` with a rank-1 I64 tensor, so `leading_broadcast(..., 1, 1)` emits nothing and `.cast(DType::I64)` is a no-op. The generated code for the two arms comes out byte-identical, so the test cannot distinguish `Shape + Tensor` from `Tensor + Shape`. A rank-3 tensor here would exercise the `unsqueeze_dims` broadcast, which is currently snapshot-only.

crates/onnx-official-tests/expectations.toml
antimora, line 3107: The `: not found in this scope` suffix is a rustc diagnostic-format difference, not a behavior change. It will churn back the next time someone regenerates on a different toolchain. Worth reverting these two rows (3107 and 3111) to keep the diff focused.

The resize and rotary-embedding reason changes further down are different: those are legitimate corrections of entries that #477 and #478 left stale. Those should stay.

antimora (commented):
Verified locally: `cargo test -p onnx-official-tests` gives 840 pass / 0 fail, and the 6 promoted `test_center_crop_pad_*_expanded` tests actually execute and compare numerically. `cargo test -p onnx-tests max` and `cargo test -p burn-onnx max` both pass. I also read the generated code in `target/.../out/model/max_{scalar,shape,shape_tensor}.rs` to confirm nothing is constant-folded away: the shape loops read `dims()` at runtime and really run.

Two things worth calling out as improvements over `sub.rs`, which should probably be back-ported: `sub.rs:81-98` passes the *other* operand's dtype into an `Int` tensor constructor (`Tensor::<1, Int>::from_data(data, (&device, DType::F32))`), and it omits rank alignment so `Tensor<1>.sub(Tensor<3>)` will not compile. This PR pins `DType::I64` and calls `leading_broadcast` first. Both are real fixes.

Three issues inline. The `as i64` one is the only one I'd consider blocking, and it is a small fix.

crates/onnx-official-tests/expectations.toml
antimora: Retracting this one, you're right. `retriage.rs:452` runs cargo with `--message-format=short`, and that format appends the primary span label after the message, so `: not found in this scope` is rustc's output, not something the PR chose. Reverting the rows would just be undone on the next `cargo xtask retriage`.

The only residual point is that the reason strings are toolchain-sensitive and will churn across rustc versions, which is a property of the tooling and not this PR's problem. Ignore the suggestion.

crates/burn-onnx/src/import/burn/node/max.rs
BenFradet: I'd rather address this one in a follow-up PR

PR #507, opened 2026-09-07
feat: OneHotEncoder support with end-to-end tests

saminiin:
**Changes**

This PR adds support for the ai.onnx.ml OneHotEncoder operator across the import pipeline.

Implemented a dedicated OneHotEncoder node processor in onnx-ir with:
- attribute extraction for cats_int64s, cats_strings, and zeros
- type inference for output dtype, rank, and static shape
- node construction and processor registration

Implemented burn-onnx code generation for OneHotEncoder:
- category table initialization as a model field
- forward generation using tensor broadcast comparison
- explicit output dtype handling

Registered OneHotEncoder in:
- onnx-ir node module and node enum mapping
- onnx-ir processor registry
- burn-onnx node module and codegen dispatch

Updated SUPPORTED-ONNX-OPS.md to mark OneHotEncoder as supported.

**Testing**

Added end-to-end integration coverage in onnx-tests:
- new OneHotEncoder model generator script
- generated ONNX fixtures for 1D, 2D, and non-default zeros attribute cases
- Rust integration tests and test/build registrations


saminiin:
@antimora Do you still review PRs or should Copilot review it first?

antimora:
> @antimora Do you still review PRs or should Copilot review it first?

I let copilot check first and look over myself.

saminiin:
@antimora Can you re-trigger Copilot to review once more. Issues it found should be fixed now.

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,319 @@
...
+                        {
+                            let x = #input.cast(burn::tensor::DType::F64);
+                            let x_unsqueezed = x.unsqueeze_dim(#input_rank);
+                            let cats = self.#field_name
+                                .clone()
+                                .cast(burn::tensor::DType::F64)
+                                .reshape([#(#ones,)* #num_categories]);
+                            x_unsqueezed.equal(cats).float().cast(burn::tensor::DType::F32)
+                        }
+                    },
```
antimora: This does not compile. The field is `Tensor<1, Int>` (see `field()` above), and `.cast()` in Burn dispatches on the input kind: `impl Cast<D, Int> for DType` has `type OutputKind = Int`, so `.cast(DType::F64)` on an Int tensor stays Int. `equal(self, other: Self)` then rejects the mismatch.

```
error[E0308]: mismatched types
  x_unsqueezed.equal(cats).float().cast(burn::tensor::DType::F32)
               ----- ^^^^ expected `Tensor<_>`, found `Tensor<2, Int>`
     = note: expected struct `burn::Tensor<_, burn::prelude::Float>`
                found struct `burn::Tensor<2, burn::prelude::Int>`
```

Verified by generating a float input model with `cats_int64s=[0,1,2,3]`, running `onnx2burn`, and compiling the output.

Use `.float().cast(DType::F64)` on the category tensor. `scaler.rs` and `svmregressor.rs` already have this shape.

There is a runtime hazard behind the type error as well: `IntDType::from(DType::F64)` panics, so `.cast(DType::F64)` on an Int tensor would abort even if the kinds lined up.

Separately, the doc on `categories()` says `cats_int64s` stay as i64 so large integer identities are not collapsed, but both of these arms cast them to F64, which collapses anything above 2^53.

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,319 @@
...
+                        {
+                            let x = #input.float().cast(burn::tensor::DType::F64);
+                            let x_unsqueezed = x.unsqueeze_dim(#input_rank);
+                            let cats = self.#field_name
+                                .clone()
+                                .cast(burn::tensor::DType::F64)
+                                .reshape([#(#ones,)* #num_categories]);
+                            x_unsqueezed.equal(cats).float().cast(burn::tensor::DType::F32)
+                        }
+                    },
```
antimora: `.float()` does not exist on a float tensor. In Burn it is defined only on `Tensor<D, Int>` (`int.rs`) and `Tensor<D, Bool>` (`bool.rs`), so the `(F32 | F64, Float64)` half of this arm fails:

```
error[E0599]: no method named `float` found for struct `burn::Tensor<1>`
  = note: the method was found for
          - `Tensor<D, burn::prelude::Bool>`
          - `Tensor<D, burn::prelude::Int>`
```

The `(I32 | I64, Float64)` half compiles. Merging the two dtype groups into one arm is what breaks it.

The `_` arm has the same problem plus the Int cast issue from the arm above. It is also unreachable: `infer_types` admits only F32/F64/I32/I64, and the four arms above cover all eight combinations. Since it is still needed for exhaustiveness over `DType`, make it emit `compile_error!` rather than duplicate a broken body.

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,319 @@
...
+                    s.parse::<f64>().unwrap_or_else(|_| {
+                        log::error!(
+                            "OneHotEncoder: cannot parse cats_strings value '{}' as a number. \
+                             String categories without numeric representation are not supported in Burn codegen.",
+                            s
+                        );
+                        0.0
+                    })
+                })
+                .collect(),
```
antimora: Falling back to 0.0 ships a silently wrong model. With `cats_strings = ["red", "green", "blue"]` all three collapse to 0.0, so the comparison tensor is `[0.0, 0.0, 0.0]`: an input of 0 sets every column (row sum 3, not one hot) and everything else gets an all zero row. Shape and dtype are correct, so nothing downstream complains.

The `log::error!` is invisible in practice. Codegen runs from `build.rs` and Cargo only surfaces build script stderr with `-vv`.

The parse diverges from the reference even when it succeeds. `onnx.reference` keys the lookup by string, so `cats_strings=["0","1","2"]` with float input `[0., 1., 2.]` returns all zeros, while this returns a full identity matrix. `test_onehotencoder_cats_strings` snapshots that wrong answer as expected.

String tensors are already rejected in `proto_conversion.rs`, so a conformant `cats_strings` model cannot reach codegen at all. Emit `compile_error!` here instead. See `compile_error_tokens` in `einsum.rs` and the usage in `deform_conv.rs`, which also has a test asserting the token.

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,319 @@
...
+        )
+    } else {
+        log::error!("OneHotEncoder: neither cats_int64s nor cats_strings provided");
+        Categories::Int64(vec![])
+    }
```
antimora: This emits `Tensor::<1, Int>::from_data([], ...)`, which does not compile:

```
error[E0282]: type annotations needed
```

Unreachable through the ONNX path today since `extract_config` rejects the neither case, but `OneHotEncoderConfig` derives `Default` and `OneHotEncoderNode::new` is public, so it is reachable through the IR API. Substituting a value that generates uncompilable code is worse than panicking with the node name.

crates/onnx-ir/src/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,359 @@
...
+                    "OneHotEncoder: exactly one of cats_int64s or cats_strings must be provided (got both)".to_string(),
+                ));
+            }
+            (None, None) => {
+                return Err(ProcessError::MissingAttribute(
+                    "cats_int64s or cats_strings".to_string(),
+                ));
+            }
+            _ => {}
+        }
```
antimora: Presence is checked, emptiness is not. `AttributeType::INTS` yields `Some(vec![])` for `cats_int64s=[]`, which passes here, gives `num_categories = 0`, and generates `from_data([], ...)` that fails with a type inference error pointing into machine generated code the user never wrote.

The reference guards this explicitly (`len(...) > 0` on both lists, then `raise RuntimeError("No encoding was defined.")`). A one line emptiness check turns a broken generated crate into a `ProcessError` naming the node.

crates/onnx-ir/src/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,359 @@
...
+                "zeros" => {
+                    if let AttributeValue::Int64(i) = value {
+                        zeros = Some(*i);
+                    } else {
+                        return Err(ProcessError::InvalidAttribute {
+                            name: "zeros".to_string(),
+                            reason: format!("expected Int64, got {value:?}"),
+                        });
+                    }
+                }
```
antimora, line 160: `zeros` is extracted and stored but never read. `grep zeros crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs` returns nothing, so the generated code always implements `zeros=1`.

Per the spec `zeros=0` means the operator must fail on an unknown category. Confirmed against `onnx.reference`: input `[0, 7, 2]` with cats `[0, 1, 2]` raises `RuntimeError: One observation did not have any defined category`, while the generated code returns an all zero row. Someone sets `zeros=0` precisely because a silent zero row is a data integrity bug in their pipeline.

Burn cannot express a data dependent abort in a pure tensor pipeline without a host sync, so rejecting `zeros=0` at codegen with a clear message is a reasonable answer. Silently downgrading it to `zeros=1` with no diagnostic is not.

Also nothing applies the spec default. `zeros` stays `None` when the attribute is absent rather than 1, which will bite whoever implements this later.

crates/onnx-ir/src/node/one_hot_encoder.rs
```diff
@@ -0,0 +1,359 @@
...
+                ));
+            }
+            _ => {}
+        }
+
+        Ok(OneHotEncoderConfig::new(cats_int64s, cats_strings, zeros))
+    }
+
+    fn build_node(&self, builder: RawNode, opset: usize) -> Node {
+        let config = self.extract_config(&builder, opset).unwrap();
```
antimora: Use `.expect("Config extraction failed")`. Every other processor in the crate does, and this is the only bare `unwrap()` in a `build_node`. `ProcessError` has a `Display` impl that `unwrap()` bypasses, so this panics with no context.

crates/onnx-tests/tests/one_hot_encoder/one_hot_encoder.py
```diff
@@ -0,0 +1,122 @@
...
+    onnx.save(model, "one_hot_encoder_2d.onnx")
+    out = ReferenceEvaluator(model).run(None, {"input": x})[0]
+    print("one_hot_encoder_2d.onnx input:\n", x)
+    print("one_hot_encoder_2d.onnx output:\n", out)
+
+
+def gen_one_hot_encoder_zeros0():
+    # Non-default attribute path: zeros=0. Inputs only contain known categories,
+    # so behavior matches zeros=1 while still exercising config extraction.
+    x = np.array([2, 1, 0], dtype=np.int64)
```
antimora: The comment is accurate, and that is the problem. Since codegen ignores `zeros`, this model is behaviorally identical to a `zeros=1` model. The test passes with `zeros=0`, with `zeros=1`, with the attribute absent, or with `zeros=99`. It reads as coverage of the non default attribute without exercising anything that distinguishes it.

`test_extract_config_reads_single_category_attribute` already asserts `zeros == Some(0)` directly and does it better.

Put an out of vocabulary value in `x` and this becomes a real test, which will then surface the divergence noted on `extract_config`.

crates/onnx-tests/tests/one_hot_encoder/mod.rs
```diff
@@ -0,0 +1,66 @@
+use crate::include_models;
+include_models!(one_hot_encoder, one_hot_encoder_2d, one_hot_encoder_zeros0);
```
antimora: All three models use INT64 input with identity categories, where the category value equals its column index. Three gaps follow from that:

1. No float input model, so the arm that does not compile is never built. Adding one fails at build time today.
2. No input containing a value outside the category list. The all zero row is the operator's defining semantic and the whole reason `zeros` exists, and it is untested for either setting.
3. No non identity categories. A bug that used the input value directly as the one hot index instead of its position in the list would pass all three tests unchanged. `cats_int64s=[10,20,30]` with input `[30,10,20]` pins the lookup.

antimora (changes requested):
The onnx-ir side is close. The codegen has a type error that makes four of the five arms in the `match (dtype, cats)` emit code that does not compile, and the one working arm is the only one with an integration test, which is why CI is green.

| arm | compiles |
| --- | --- |
| `(I32 or I64, Int64)` | yes |
| `(F64, Int64)` | no, E0308 |
| `(F32, Int64)` | no, E0308 |
| `(F32 or F64, Float64)` | no, E0599 |
| `_` | no, E0599, and dead |

Any float typed input produces a model that fails to build. `infer_types` accepts `tensor(float)` and `tensor(double)` and the module doc advertises both, so this is not an edge case, it is the common sklearn to ONNX shape.

Verified by generating float input and cats_strings models, running `onnx2burn`, and compiling the emitted code against the pinned burn rev.

Two behavior issues on top of that. `zeros=0` silently returns an all zero row where ONNX raises. `cats_strings` is parsed to f64, which gives a different answer from `onnx.reference` when the parse succeeds and a silently wrong model when it fails.

Blocking: the type fix plus a float input integration test. The rest can follow.

Minor: `onnx-spec/ops/OneHotEncoder.md` was not added. `Imputer.md`, `Scaler.md`, and `SVMRegressor.md` are all there.

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
saminiin: Resolved by splitting the dtype handling into separate float and integer arms;

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
saminiin: Fixed, no more returning 0.0

crates/burn-onnx/src/import/burn/node/one_hot_encoder.rs
saminiin: This code doesn't exist anymore

crates/onnx-ir/src/node/one_hot_encoder.rs
saminiin: Fixed in validation. Empty cats_int64s now returns a ProcessError before codegen, and a regression test covers cats_int64s=[]; it no longer fails later in generated code.

crates/onnx-ir/src/node/one_hot_encoder.rs
saminiin: Zeros now defaults to 1 when absent (unwrap_or(1)), and any non-1 value (including zeros=0) is rejected with a ProcessError before codegen.

crates/onnx-ir/src/node/one_hot_encoder.rs
saminiin: Fixed.

PR #518, opened 2026-09-11
fix: handle shape broadcast correctly

BenFradet:
### Checklist

### Related Issues/PRs

- stemmed from #503 
- convo https://github.com/tracel-ai/burn-onnx/pull/503#discussion_r3930650344

### Changes

This fixes shape/shape interactions whenever broadcast is needed

### Testing

- new insta tests
- new onnx conformity tests


crates/onnx-tests/build.rs
```diff
@@ -48,59 +48,59 @@ fn add_all_inputs(model_gen: &mut ModelGen) {
         .input("tests/acos/acos.onnx")
         .input("tests/acosh/acosh.onnx")
         .input("tests/add/add.onnx")
-        .input("tests/asin/asin.onnx")
-        .input("tests/asinh/asinh.onnx")
-        .input("tests/atan/atan.onnx")
-        .input("tests/atanh/atanh.onnx")
-        .input("tests/add/add_shape.onnx")
+        .input("tests/add/add_argmax_with_shape.onnx")
```
BenFradet, line 51: unsorted list was stressing me out

crates/burn-onnx/src/import/burn/node/broadcast_helpers.rs
```diff
@@ -32,6 +32,50 @@ pub(crate) fn leading_broadcast(
...
+        {
+            let __lhs = #lhs;
+            let __rhs = #rhs;
+            let mut __result = [0i64; #len_lit];
+            #[allow(clippy::needless_range_loop)]
+            for __i in 0..#len_lit {
+                __result[__i] = #elem;
+            }
+            __result
+        }
```
antimora, line 70: Looks ugly; can we improve the code generation?

BenFradet: moved to `array::from_fn`

crates/burn-onnx/src/import/burn/node/add.rs
```diff
@@ -260,11 +260,59 @@ mod tests {
...
-                    *result_item = result_item.saturating_add(*rhs_item);
-                }
-                result
+                let __lhs = lhs;
+                let __rhs = rhs;
+                core::array::from_fn::<
+                    i64,
+                    4usize,
+                    _,
+                >(|__i| (__lhs[__i]).saturating_add(__rhs[__i]))
```
antimora, line 269: can we also not use the local type of variables like `__lhs` ? The variables are already enclosed in a block so there isn't any name collision. Typically the `__a` type of reassignments are to re-declare as `mut` or changing types.

antimora: It's probably the existing code base telling stylistic hints to AI agents. So I am fixing this and I'll add new rules to AGENTS.md not to generate code like that.

BenFradet: yes, tbh I copied the approach from the other code in `broadcast_helpers`

Issue #311, opened 2026-04-08
ONNX backend test suite: 181 failures across reduce/cast/resize/nllloss op families (90.1% pass)


antimora:
## Summary

Running the full ONNX backend test suite against burn-onnx (via the scoreboard integration drafted in #299) reveals 181 failing tests out of 1827 CPU tests. Pass rate: **1646/1827 (90.1%)**. Runtime: 21 minutes.

Failures are concentrated in specific op families, not the integration layer itself. This issue tracks the correctness gaps that need to be addressed before submitting to the ONNX Backend Scoreboard.

## Test Environment

- Branch: `issue/299-onnx-backend-scoreboard`
- Backend wrapper: `scoreboard/backend.py`
- Runner: `scoreboard/runner/` (burn + burn-ndarray, NdArray<f32> backend)
- onnx2burn flags: `--no-development --no-simplify`
- Full raw log: https://gist.github.com/antimora/2178ee7beddfe57d5cda5f3edc5aeb63
- Full failing test list: https://gist.github.com/antimora/a0ee7679550aa632c671ef40412e5359

## Failures by Op Family

| Op Family | Count | Root Cause (preliminary) |
|---|---:|---|
| `reduce_sum` | 24 | Output shape differs (keepdims/rank handling) |
| `castlike` | 18 | Exotic dtypes: FLOAT4E2M1, FLOAT8E*, INT2/4, UINT2/4 |
| `reduce_l1` | 14 | Output shape (keepdims, empty set) |
| `reduce_l2` | 14 | Output shape (keepdims, empty set) |
| `nllloss` | 12 | Weight handling, reduction modes |
| `resize` | 11 | Algorithm differences (cubic, nearest modes) |
| `cast` | 10 | Same exotic dtypes as castlike |
| `reduce_log_sum` | 9 | Axis handling, empty set |
| `attention` | 8 | Numerical precision on 3D/4D with causal/bias/mask |
| `reduce_max` | 7 | Output shape + bool inputs |
| `reduce_min` | 7 | Output shape + bool inputs |
| `reduce_prod` | 7 | Output shape (keepdims) |
| `reduce_mean` | 6 | Output shape (keepdims) |
| `triu` | 6 | Negative diagonal, edge cases |
| `tril` | 5 | Negative diagonal, edge cases |
| `averagepool` | 4 | `ceil_mode` with same padding |
| `bernoulli` | 4 | Random op (may need seed passthrough) |
[...]

PR #523, opened 2026-09-14
Handle negative indices in Gather and GatherElements

yuefdev:
### Checklist

- [x] Confirmed that `cargo xtask validate` command has been executed. Audit, format and lint pass. Locally it then stops at the typos step on `crates/onnx-ir/src/node/einsum.rs:990` (`"ij,jk->iz"`, an intentionally invalid equation in an existing test), which this PR doesn't touch, so I ran the relevant test suites directly (see Testing).
- [x] Confirm relevant documentation has been updated. No docs change needed.
- [x] For new/modified ONNX operators: verified implementation matches `onnx-spec/ops/<OpName>.md`. Both specs allow indices in `[-s, s-1]` along the gather axis.

### Related Issues/PRs

- Promotes `test_gather_negative_indices` and `test_gather_elements_negative_indices` from `fail-compare` to `pass` (both were tracked under #311)
- Follows the approach ScatterElements already uses for negative indices (#487)

### Changes

Gather and GatherElements passed the index tensor straight to `take` / `gather`. Those only accept non-negative indices, so a model with runtime negative indices got wrong values.

The generated code now folds negative indices first by adding the axis size where the index is below zero (`lower_elem` + `mask_where`), then calls `take` / `gather` as before. Out-of-range indices stay unchecked, as the spec makes them an error and ScatterElements doesn't check them either.

Only the tensor-index path of Gather changes. The Shape paths already normalise negatives, and scalar indices go through `s![]`, which handles them.

### Testing

- `cargo test -p burn-onnx --lib gather`: 41 passed (5 inline snapshots updated for the new prologue)
- `cargo test -p onnx-official-tests`: 854 passed, including the two promoted tests


crates/burn-onnx/src/import/burn/node/gather.rs
```diff
@@ -327,8 +327,22 @@ fn forward_tensor_gather(
...
+                        let #output = {
+                            let __gather_input = #input;
+                            let __gather_axis_size = __gather_input.dims()[#dim] as i64;
+                            let __gather_indices = #index;
+                            let __gather_negative = __gather_indices.clone().lower_elem(0i64);
+                            let __gather_corrected = __gather_indices.clone() + __gather_axis_size;
+                            let __gather_indices =
+                                __gather_indices.mask_where(__gather_negative, __gather_corrected);
+                            __gather_input
+                                .take::<#index_rank_lit, #final_rank_lit>(#dim, __gather_indices)
```
antimora: Please rebase from the main. Just recently I added a rule in Agents.md not to use `__`

antimora (approved):
LGTM!

Thank you for your contribution!

